PR #18, opened 2022-12-01
Upgrade pgx to 0.6.0

yrashk:
Problem: wrappers uses pgx 0.5.6

This version is outdated.

Solution: upgrade to 0.6.0

I've also set the default version of Postgres to 15 as pgx supports it now and this is the most recent version that keeps evolving, therefore it would be better to track it to catch new failures.

Dependency update

Using pgx 0.5.6

## What is the new behavior?

Using pgx 0.6.0

### DIFF
--- a/wrappers/bin/installcheck
+++ b/wrappers/bin/installcheck
@@ -25,7 +25,7 @@ rm -rf "$tmpdir"
 # Initialize: setting PGUSER as the owner
 initdb --no-locale --encoding=UTF8 --nosync -U "$PGUSER"
 # Start the server
-pg_ctl start -o "-F -c listen_addresses=\"\" -c log_min_messages=WARNING -c shared_preload_libraries='pgsodium' -c pgsodium.getkey_script='/usr/share/postgresql/14/extension/pgsodium_getkey.sh' -k $PGDATA"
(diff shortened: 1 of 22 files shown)


wrappers/bin/installcheck
```diff
@@ -25,7 +25,7 @@ rm -rf "$tmpdir"
 # Initialize: setting PGUSER as the owner
 initdb --no-locale --encoding=UTF8 --nosync -U "$PGUSER"
 # Start the server
-pg_ctl start -o "-F -c listen_addresses=\"\" -c log_min_messages=WARNING -c shared_preload_libraries='pgsodium' -c pgsodium.getkey_script='/usr/share/postgresql/14/extension/pgsodium_getkey.sh' -k $PGDATA"
+pg_ctl start -o "-F -c listen_addresses=\"\" -c log_min_messages=WARNING -c shared_preload_libraries='pgsodium' -c pgsodium.getkey_script='/usr/share/postgresql/15/extension/pgsodium_getkey.sh' -k $PGDATA"
```
olirice, line 28: any way this could be parameterized? The tests won't run against 14 if we hard code it

olirice (commented):
lgtm

olirice (approved):

wrappers/bin/installcheck
yrashk: We can do this as a part of a CI overhaul

PR #20, opened 2022-12-03
Problem: extensive use of macros in supabase-wrappers

yrashk:
There are at least two problems with this:

1. Development experience

Most of the code is in macros and becomes hard to reason about and navigate using IDEs as it is all simply quoted expressions.

2. One macro approach captures all provided FDWs and will conflict

"Magic" macro provides pg_extern function that exposes all listed FDWs in one bundle. However, should any other extension use supabase-wrappers, it won't install alongside `wrappers` extension because of a duplicate symbol.

3. Compile times may suffer with macro-heavy code.

Solution: move all this code to supabase-wrappers

As my analysis shown, there's no real need for macros. In fact, in some parts (like use of ForeignDataWrapper::new without it being part of the trait) are downright wrong.

For the time being, I've removed the "meta" listing functionality as it would need to be re-designed to support multiple extensions and/or unbundled packaging of multiple FDWs. If it will be deemed necessary, I can see a number of ways it can re-implemented.

There's a lot that can be improved, but I've tried to keep this change short and to the point (and that point is removing the dependency on macros)

Architecture improvement

Relying on a huge macro

## What is the new behavior?

Not relying on a huge macro


supabase-wrappers/src/interface.rs
```diff
@@ -312,4 +317,48 @@ pub trait ForeignDataWrapper {
...
+    }
+
+    /// Additional FwdRoutine setup, called by default `Self::fdw_routine()`
+    /// after completing its initialization.
+    fn fdw_routine_hook(_routine: &mut FdwRoutine<AllocatedByRust>) {}
+
+    fn validator(_options: Vec<Option<String>>, _catalog: Option<pg_sys::Oid>) {}
+}
+
+pub enum ValidatorScope {
```
burmecia, line 362: What is this `ValidatorScope` for? Cannot see it is used anyware.

yrashk: Oh. I was thinking about further evolution and forgot to remove it. I'll get to my computer and remove it from the PR.

yrashk: Updated.

burmecia:
That's amazing work! Thanks @yrashk.

The initial thought was to use a single handler and trait objects to dynamically create fdw instances, and also to provide a 'centralised' fdw meta view. It looks like this 'distributed' approach is better than the macro-based approach. And you are right the macro is really a pain when debugging so I am happy to see it can be removed.

yrashk:
Updated to resolve conflicts.

yrashk:
> That's amazing work! Thanks @yrashk.
> 
> The initial thought was to use a single handler and trait objects to dynamically create fdw instances, and also to provide a 'centralised' fdw meta view. It looks like this 'distributed' approach is better than the macro-based approach. And you are right the macro is really a pain when debugging so I am happy to see it can be removed.
> 
> Aslo after removing the macro, the `validator` function is unlocked and can be customised by each fdw, that's another benefit, great!

Yeah, that's very much my thinking as well. And I absolutely go crazy navigating and debugging macros, they are so much pain. That's why I reserve their use to situations where there's nothing better. Clearly there's something better available here.

And as for centralized registry of FDWs, we can still achieve it. For example: creating a meta function for every FDW that returns meta information. It has to follow a certain pattern to be picked up by meta "scanner". Or some variation of it. Personally, I am not sold on the necessity of such a feature at this time anyway -- I don't see much value (but I may be missing something)

burmecia:
> And as for centralized registry of FDWs, we can still achieve it. For example: creating a meta function for every FDW that returns meta information. It has to follow a certain pattern to be picked up by meta "scanner". Or some variation of it. Personally, I am not sold on the necessity of such a feature at this time anyway -- I don't see much value (but I may be missing something)
Agree, a centralised meta view seems no need, but a kind of meta information still needed to provide basic info to the fdw users, especially when upgrading. Maybe we can add a meta function in the `ForeignDataWrapper` trait to ask each fdw to provide it in an unified format?

yrashk:
> > And as for centralized registry of FDWs, we can still achieve it. For example: creating a meta function for every FDW that returns meta information. It has to follow a certain pattern to be picked up by meta "scanner". Or some variation of it. Personally, I am not sold on the necessity of such a feature at this time anyway -- I don't see much value (but I may be missing something)
> 
> Agree, a centralised meta view seems no need, but a kind of meta information still needed to provide basic info to the fdw users, especially when upgrading. Maybe we can add a meta function in the `ForeignDataWrapper` trait to ask each fdw to provide it in an unified format?

That's very easy to add, indeed. This is outside of the scope of this PR.

While outside of the scope of this PR, I'd like to further elaborate on how I see FDW exposure improving:

Currently exposing a handler and a validator require the manual creation of two very tiny `pg_extern` functions simply calling functions in the trait. While they are not bothering me too much, I think this can be further improved by having a macro that does that automatically.

It can be either a simple syntax macro (thus requiring no additional crate):

```rust
fdw_wrapper!(HelloWorldFdw, hello_world, meta-information);
```

Or it can be a proc macro (that would require a proc-macro crate again, but a tiny one):

```
#[fdw_wrapper]
#[meta(...)]
pub struct HelloWorldFdw {
}
```

Either of them will simple expose a few small `pg_extern` functions, and that's it.

Again, I suggest we separate that work from this as the purpose of this PR is to simply remove that monster macro so we can continue moving ahead with better velocity.

wrappers/src/fdw/airtable_fdw/mod.rs
```diff
@@ -1,4 +1,17 @@
+use supabase_wrappers::prelude::*;
 mod airtable_fdw;
 mod result;
+use pgx::pg_sys;
+use pgx::prelude::*;
 
 pub(crate) use airtable_fdw::AirtableFdw;
+
+#[pg_extern]
+fn airtable_fdw() -> supabase_wrappers::FdwRoutine {
```
burmecia, line 10: These two functions to provide handler and validator seems repetitive, can we use a macro like `setup_fdw()` to simplify it further? Also can we set its name pattern like `xxx_fdw_handler()` and `xxx_fdw_validator()`, so that it is easy to filter when the wrappers extension is installed with others in a same schema.

burmecia:
okay, I will merge this PR and thanks again!

burmecia (approved):

wrappers/src/fdw/airtable_fdw/mod.rs
yrashk: They are repetitive, and I do think having a simple "registration" macro is useful. Either as a simple macro or as a type-level macro. I mentioned this in the PR. We will get this fixed easily.

PR #29, opened 2022-12-07
Problem: CI is painfully slow

yrashk:
Solution: use a larger runner (4 cores, 16 GB RAM)


olirice:
doesn't look like we have access to those runners so I canceled the workflow.

Lets revisit this another time when everyone is less busy

PR #30, opened 2022-12-08
chore: add more Stripe test cases

burmecia:

This PR is to add more smoke test cases for Stripe FDW.

Currently, there are only 3 test cases:

1. Balance
2. Customers
3. Subscriptions

## What is the new behavior?

More test cases will be added:

1. Balance
2. Balance Transactions
3. Charges
4. Customers
5. Invoices
6. PaymentIntents
7. Products
8. Subscriptions

## Additional context

N/A


olirice (changes requested):
looks great!

for the tables that support insert, update, and where push down could that included too?

olirice (approved):

PR #37, opened 2022-12-14
fix: change memory context parent to CurTransactionContext

burmecia:

This PR is to change temporary memory context's parent to `CurTransactionContext`, which will fix crash issue when using some SQL client tools, such as DBeaver.

Some SQL client tools, like DBeaver, will change the default `CurrentMemoryContext` which may cause FDW crash as the state structure and its temporary memory context's using `CurrentMemoryContext` as their parent context.

## What is the new behavior?

The state structure and its temporary memory context will be fixed to allocate under `CurTransactionContext`.

## Additional context

So far, only DBeaver has this issue, psql and Supabase are not affected.

### DIFF
--- a/supabase-wrappers/src/modify.rs
+++ b/supabase-wrappers/src/modify.rs
@@ -37,7 +37,8 @@ impl<W: ForeignDataWrapper> FdwModifyState<W> {
             rowid_attno: 0,
             rowid_typid: 0,
             opts: HashMap::new(),
-            tmp_ctx: PgMemoryContexts::new("Wrappers temp modify data"),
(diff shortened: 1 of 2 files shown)


soedirgo (approved):

yrashk:
Do you know what is DBeaver doing differently that causes this issue, in terms of operating its connection (as it doesn't use PG's internals to set `CurrentMemoryContext)?

burmecia:
Not sure what DBeaver did when connect, but sometimes I saw `Assert(context != CurrentMemoryContext)` failure. I suspect that may be related to `CurrentMemoryContext`, so I changed all the memory to `CurTransactionContext` and that failure disappeared.

PR #40, opened 2022-12-15
Add Array/Object type handling to Airtable wrapper

csmithxc:

Introduces feature that allows Airtable FDW to handle `Object` and `Array` types from Airtable API.

FDW is not currently able to map Airtable `Object` / `Array` types to postgres. If it is attempted, the FDW handler raises a panic.

## What is the new behavior?

Airtable `Object` and `Array` types are mapped to `pgx::JsonB`. This means that these types must be mapped to a `jsonb` postgres column data type.

## Additional context

These types are especially useful, as this now allows users to handle Airtable types such as:
 - Image
 - Link to another record

This is by no means bulletproof (I am a Rust noob) or necessarily able to handle every Airtable type that maps to array / object, but it is certainly better than not being able to handle any of these Airtable types at all.

### DIFF
--- a/wrappers/src/fdw/airtable_fdw/result.rs
+++ b/wrappers/src/fdw/airtable_fdw/result.rs
@@ -1,3 +1,4 @@
+use pgx::JsonB;
 use serde::de::{MapAccess, Visitor};
 use serde::{Deserialize, Deserializer};
 use serde_json::Value;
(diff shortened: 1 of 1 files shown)


elvise:
Hi @ankrgyl can you review it ?

ankrgyl:
I think someone from Supabase (CC @burmecia) should drive the review

elvise:
@burmecia just a gentle reminder :)

burmecia (approved):
Thanks for the PR!

PR #46, opened 2022-12-22
Perf improvements

EdMcBane:

A new iter_scan method to reuse a borrowed `Row`; a few additional performance improvements.

`ForeignDataWrapper::iter_scan` requires the implementor to build a Row, which is quite expensive.

## What is the new behavior?

`ForeignDataWrapper::iter_scan_borrowed` takes a mutable ref to a Row owned by the caller; the implementor just needs to populate it.
A default implementation is provided that simply delegates to `ForeignDataWrapper::iter_scan` and swaps the returned `Row`.

## Additional context

On my machine (intel 8700k), in release mode, pg15, with a sample FDW that yields 10M "helloworld" rows, the time taken to execute `SELECT * FROM bench WHERE id < 0` are:
* Original: 3244,732 ms (00:03,245)
* Without `debug2!` on `iter_scan`: 1565,417 ms (00:01,565)
* Reusing `Row`: 1241,977 ms (00:01,242)

I've been unable to run the tests on this, the docs seem out of date and `cargo test` fails to compile.

### DIFF
--- a/supabase-wrappers/src/interface.rs
+++ b/supabase-wrappers/src/interface.rs
@@ -189,6 +189,11 @@ impl Row {
         iter = keep.iter();
         self.cells.retain(|_| *iter.next().unwrap());
     }
+
+    pub fn clear(&mut self) {
+        self.cols.clear();
+        self.cells.clear();
+    }
 }
 
 /// A restiction value used in [`Qual`], either a [`Cell`] or an array of [`Cell`]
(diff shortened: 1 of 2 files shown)


EdMcBane:
Example wrapper used for "benchmarks"

https://github.com/EdMcBane/wrappers_bench

burmecia:
That's great improvement,  thanks @EdMcBane ! I think this should be the default interface and we will make it in the next release.

burmecia (approved):

PR #48, opened 2022-12-28
chore: use borrowed row in iter_scan()

burmecia:

This PR is to use borrowed `row` in `iter_scan()` to improve performance, see #46 .

Currently in each call of the `iter_scan()` function, an empty new row needs to be created and values are copied over to Postgres, this can be slow when reading large number of rows.

## What is the new behavior?

The new `iter_scan()` function will pass a pre-allocated `row` and FDW can optionally use `mem::replace()` to directly move data in.

## Additional context

See more details in #46 .

### DIFF
--- a/supabase-wrappers/src/interface.rs
+++ b/supabase-wrappers/src/interface.rs
@@ -10,6 +10,7 @@ use pgx::{
 use std::collections::HashMap;
 use std::fmt;
 use std::iter::Zip;
+use std::mem;
 use std::slice::Iter;
 
 // fdw system catalog oids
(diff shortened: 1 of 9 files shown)


olirice:
@burmecia I took a look but memory safety stuff is over my head. I pinged @yrashk to take a look when he gets a sec

concept sounds great!

yrashk:
No safety issues pop out for me here, but I think there may be further room for improvement. `replace_with` is still a byte copy of Row. Rows are still being created, moved, and copied.

I'd suggest establishing a benchmark test and iterating over this design further. Happy to assist there.

burmecia:
The byte copy of row is needed anyway, the saved cost is actually the repetitive row creation and data movement from fdw instance to `supabase-wrappers`, use borrowed row can remove that.

olirice (approved):

PR #55, opened 2023-02-05
:sparkles: Support Stripe Connect

arnaud:
Support for accounts and transfers


burmecia (approved):
Thank you for your PR!

Issue #58, opened 2023-02-17
`qual.deparse()` null check is incorrect


burmecia:
# Bug report

## Describe the bug

The `qual.deparse()` incorrectly treats null check conditions.

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

1. for any query with null check in condition like `select x from foo where x is null`
2. the `qual.deparse()` will return string like `x is 'null'`

## Expected behavior

The null check qual should be deparsed correctly like `x is null`

## Screenshots

N/A

## System information

- OS: any
- Version of wrappers: v0.1.7
- Affected FDWs: any

## Additional context

N/A

PR #59, opened 2023-02-17
fix: add special treatment for null check in qual

burmecia:

This PR is to fix a bug in `qual.deparse()` when dealing with null check condition.

#58

## What is the new behavior?

The null check in qual should be deparsed correctly.

## Additional context

N/A


olirice (approved):
@burmecia approved! would be great to add a test for this too so it can't regress

Issue #70, opened 2023-03-10
BigQuery wrapper: JSON, Numeric not supported


yan-hic:
Trying to query a table with JSON data type in BigQuery.

Foreign table, with `json` column type, gets created fine but upon querying, the wrapper returns:
```
Failed to run sql query: get table metadata failed: Request error (error: error decoding response body: unknown variant `JSON`, expected one of `STRING`, `BYTES`, `INTEGER`, `INT64`, `FLOAT`, `FLOAT64`, `NUMERIC`, `BIGNUMERIC`, `BOOLEAN`, `BOOL`, `TIMESTAMP`, `DATE`, `TIME`, `DATETIME`, `RECORD`, `STRUCT` at line 15 column 22)
```

yan-hic:
The workaround I used till supported: 

1. use `json_to_string()` in bigquery
2. in the foreign table ddl, define that column as `text` and add a generated column to convert back to json (`jsonStringColumn::json`)

burmecia:
Yes, JSON is not support yet in BigQuery FDW. A workaround is using `TO_JSON_STRING` to convert JSON to STRING in BigQuery and store that string as source table column. Then use subquery or view on top of PG foreign table and convert that string back to JSON.

burmecia:
We're currently working on improving BigQuery fdw, the JSON support should be ready on the next release. And thanks for the subquery in `table` option, it is quite useful, we will evaluate and hopefully include it in the future release.

yan-hic:
EDIT: I have edited the title and added `Numeric`:
`Failed to run sql query: field type Numeric not supported`

which is a valid type in PG: https://www.postgresql.org/docs/current/datatype-numeric.html

Would be good to add all supported ones or document the limited support.
Thanks.

rubenvereecken:
@burmecia I believe this should be reopened. #78 added numeric support, but not JSON, which was the main issue raised originally.

jscheel:
@burmecia what happened to the json support?

burmecia:
@jscheel, it has JSON writing support for now, but not for reading. We will be adding it soon.

jscheel:
@burmecia awesome! Thank you!

PR #74, opened 2023-03-15
feat: add subquery to bq fdw table option

### LINKED ISSUE #69: BigQuery wrapper: all columns must be defined
The BQ wrapper works great out-of-the-box on supabase cloud. A game changer really for us as we only started evaluating SB yesterday.

One word of caution though - or a feature request ?: when creating the PG foreign table as per the instructions, the list of columns must match exactly the source table. Ideally one would want to define any subset instead i.e. if you only need the read a few fields. If no plan to change, please update the documentation ( + also update https://supabase.com/docs/guides/database/extensions/wrappers and  https://supabase.github.io/wrappers/ as BigQuery is available)

Workaround (especially if lots of columns): create & use a BQ view as source.

burmecia:

This PR is to add subquery support to `table` option in BigQuery FDW, possibly fix #69 and #70 . 

Currently, the `table` foreign table option in BigQuery FDW can only specify a table or view name, which limits the fdw usage in some cases.

## What is the new behavior?

In the `table` option we can specify a subquery enclosed in parentheses. For example,

```sql
OPTIONS (
    table '(select * except(props), to_json_string(props) as props from `my_project.my_dataset.my_table`)'
)
```

## Additional context

As the [dependent lib](https://docs.rs/gcp-bigquery-client/latest/gcp_bigquery_client/model/field_type/enum.FieldType.html) doesn't support JSON type at this moment, this feature can be used as a workaround to read JSON data from BQ.

### DIFF
--- a/wrappers/src/fdw/bigquery_fdw/README.md
+++ b/wrappers/src/fdw/bigquery_fdw/README.md
@@ -91,7 +91,7 @@ create foreign table people (
   options (
     table 'people',     -- source table in BigQuery, required
                         -- This can also be a subquery enclosed in parentheses, for example,
-                        -- `table '(select * except(props), to_json_string(props) as props from ``my_project.my_dataset.my_table``)`
+                        -- table '(select * except(props), to_json_string(props) as props from `my_project.my_dataset.my_table`)'
     location 'EU',      -- table location, optional
     rowid_column 'id',  -- primary key column name, optional for scan, required for modify
     startup_cost '42'   -- execution startup cost for exection planning, optional
--- a/wrappers/src/fdw/bigquery_fdw/README.md
+++ b/wrappers/src/fdw/bigquery_fdw/README.md
@@ -90,6 +90,8 @@ create foreign table people (
   server my_bigquery_server
   options (
     table 'people',     -- source table in BigQuery, required
+                        -- This can also be a subquery enclosed in parentheses, for example,
+                        -- `table '(select * except(props), to_json_string(props) as props from ``my_project.my_dataset.my_table``)`
     location 'EU',      -- table location, optional
     rowid_column 'id',  -- primary key column name, optional for scan, required for modify
     startup_cost '42'   -- execution startup cost for exection planning, optional
@@ -133,5 +135,6 @@ insert into people values (4, 'Yoda', current_timestamp());
 
 | Version | Date       | Notes                                                |
 | ------- | ---------- | ---------------------------------------------------- |
-| 0.1.0   | 2022-11-30 | Initial version                                      |
+| 0.1.2   | 2023-03-15 | Added subquery support for `table` option            |
 | 0.1.1   | 2023-02-15 | Upgrade bq client lib to v0.16.5, code improvement   |
+| 0.1.0   | 2022-11-30 | Initial version                                      |
--- a/wrappers/src/fdw/bigquery_fdw/bigquery_fdw.rs
+++ b/wrappers/src/fdw/bigquery_fdw/bigquery_fdw.rs
@@ -83,7 +83,7 @@ fn field_to_cell(rs: &ResultSet, field: &TableFieldSchema) -> Option<Cell> {
 }
 
 #[wrappers_fdw(
-    version = "0.1.1",
+    version = "0.1.2",
     author = "Supabase",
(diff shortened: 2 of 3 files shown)


olirice (commented):
Since this solution would not work with updates and deletes I opened https://github.com/lquerel/gcp-bigquery-client/issues/56 to see if upstream support for the JSON data type is possible

If so, that might be the way to go so we can retain update/delete with json columns

olirice (approved):
Approved, but lets wait to see if we get an upstream response about adding support in the driver before merging

yan-hic:
Merged and I see the doc has been updated so I guess it's GA, I am getting this error in supabase cloud: 
```
Failed to run sql query: get table metadata failed: Response error (error: ResponseError { error: NestedResponseError { code: 400, errors: [{"domain": "global", "message": "Invalid table ID \"(select * from inv_taxtract)\".", "reason": "invalid"}], message: "Invalid table ID \"(select * from inv_taxtract)\".", status: "INVALID_ARGUMENT" } })
```
Note that I have recreated fwd, server and foreign table, to no avail. Should the extension be disabled and reenabled again to load the latest version ?


Also, the doc mentions `table '(select * except(props), to_json_string(props) as props from `my_project.my_dataset.my_table`)'
` as an example but the project and dataset should be removed since the fdw server defines them already. I assume they cannot be different.

burmecia:
Yes, the BigQuery FDW is GA now. Could you upgrade to latest version 0.1.9 and try again? You can run `select * from pg_catalog.pg_extension where extname = 'wrappers'` to check the installed version. 

The custom sql in `table` option must specify full table qualifiers like `my_project.my_dataset.my_table`, because it is user specified sql, the project and dataset defined on fdw server are not used.

yan-hic:
@burmecia I can't figure how to get the latest version. here the steps I have been through:

1.  deleted all dependent objects
2.  removed (toggle off) the 'WRAPPERS' extension
3. confirmed it's gone: `select * from pg_catalog.pg_extension where extname = 'wrappers' ` empty
4. re-enabled (toggle on)
5. `select * from pg_catalog.pg_extension where extname = 'wrappers' ` still shows 0.1.7

Is this supposed to be updated automatically for those having the wrappers enabled already (and is a bug then) - or what are the steps to follow ? In the latter case, would be good to document publicly.

burmecia:
The latest version of wrappers comes with new Supabase projects, it won't automatically upgrade for existing projects. Also you need to manually enable it by toggling on in Studio page or using sql as below. You probably need to upgrade to the latest postgres image first (see 'Infrastructure' -> 'Postgres' on https://app.supabase.com/project/_/settings/general page), and then can use latest version of wrappers.

Some useful queries:

1. list all available versions: `SELECT * FROM pg_catalog.pg_available_extension_versions WHERE name = 'wrappers'`
2. check installed version: `select * from pg_catalog.pg_extension where extname = 'wrappers'`
3. install wrappers extension: `create extension wrappers`
4. upgrade extension: `alter extension wrappers update to '0.1.9'`  (version number must be in the available versions)

yan-hic:
There is no upgrade or migration option under infrastructure, or anywhere else. Disappointing that extension updates are only available to new projects. In most cases, that would mean taking a PG backup and starting over.
No worth effort so we will stick to it until there is a better/easier path.

yan-hic:
FYI - solved after https://github.com/orgs/supabase/discussions/13425 !

yan-hic:
Well, I guess I am still unclear on the release pipeline. 
I followed instruction to pause/restart project from Gary. That got me 0.1.9, but not 0.1.10 which was released yesterday (for the numeric support). 
What else should happen for the latest release to be available on supabase cloud/free ? Something on your end ?

burmecia:
Wrappers v0.1.10 was just released, it needs to be included the [postgres image](https://github.com/supabase/postgres) which is used by Supabase platform. That process may take some time, as the changes need reviews and be progressively rolled out.

Here is [the PR for v0.1.10 upgrading](https://github.com/supabase/postgres/pull/608) in postgres image, after that PR is merged to master then it will soon be ready for new projects (or upgrading existing projects) on Supabase platform.

yan-hic:
Gotcha @burmecia. Will make sure to wait for the supa PR merge in the future. Thanks again !

PR #84, opened 2023-05-01
Feat/stripe pricing

nielsbrakel:

This PR has two changes

- Fixed typos in the docs
- Added new prices connected object

Currently you are only able to get the product but not the prices of the products.

## What is the new behavior?

You can now access the prices and use them in a pricing page for example.

## Additional context

I was not able to get it running on my local machine so hopefully the everything is green 🙃

### DIFF
--- a/docs/firebase.md
+++ b/docs/firebase.md
@@ -56,7 +56,7 @@ create server firebase_server
 
 #### Auth (Insecure)
 
-If the platform you are using does not support `pgsodium` and `Vault`, you can create a server by storing yourt [service account key](https://firebase.google.com/docs/admin/setup#add_firebase_to_your_app) directly.
+If the platform you are using does not support `pgsodium` and `Vault`, you can create a server by storing your [service account key](https://firebase.google.com/docs/admin/setup#add_firebase_to_your_app) directly.
 
 
 !!! important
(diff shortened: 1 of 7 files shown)


burmecia (approved):

burmecia:
thank you very much for the PR!

nielsbrakel:
@burmecia Thanks for merging my PR (their might be more coming for other stripe objects). I just tried to add the foreign table but it does not work yet.

image

this also might be a bug that it does not need to be called `pricing` but `prices` but that resulted in the same error:

image

Could you provide me with some insights on how the release process works? This will greatly help me in making this wrapper more complete.

burmecia:
@nielsbrakel , the release process is a bit long, basically it has to be included in Supabase platform to use:

1. release a version of Wrappers
2. include the new Wrappers in [Supabase Postgres](https://github.com/supabase/postgres) 
3. Supabase Postgres release a new version to Docker or Supabase platform
4. Once it is released on Supabase platform, new project will have the new version immediately, but existing projects will wait for another another round of roll out which needs more time

So the whole process may take several weeks, if you are eager to use the new version, you can raise a support ticket and then we will see if we can help.

PR #89, opened 2023-05-06
Feat/Stripe: Added primary keys to all objects and small docs fix

nielsbrakel:

This PR has two changes

- Fixed typo in docs for create script of `prices` object
- Added `rowid_column` to all objects  containing a `id`

Currently there are some columns that have a `rowid_column` and other who don't. For consistency I added it to all objects containing an `id`. As far as I understands this is not the same as a real constraint because in a foreign table there is no way for PostgreSQL to enforce this behavior but I still think it could help with being a simple form of documentation of the code. Hopefully my hypothesis is correct, and otherwise I'm curious to the explanation since I'm here to learn.

## What is the new behavior?

- All rows with an id now have the `rowid_column` added.

## Additional context

https://www.postgresql.org/docs/current/sql-createforeigntable.html#notes

I was not able to get it running on my local machine so hopefully everything is green 🙃

### DIFF
--- a/docs/stripe.md
+++ b/docs/stripe.md
@@ -112,7 +112,8 @@ create foreign table stripe.accounts (
 )
   server stripe_server
   options (
-    object 'accounts'
+    object 'accounts',
+    rowid_column 'id'
   );
 ```
 
@@ -162,7 +163,8 @@ create foreign table stripe.balance_transactions (
 )
   server stripe_server
   options (
-    object 'balance_transactions'
+    object 'balance_transactions',
+    rowid_column 'id'
   );
 ```
 
@@ -193,7 +195,8 @@ create foreign table stripe.charges (
 )
   server stripe_server
   options (
-    object 'charges'
+    object 'charges',
+    rowid_column 'id'
   );
 ```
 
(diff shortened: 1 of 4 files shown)


burmecia:
Hi @nielsbrakel , thank you for your PR! The reason why some objects don't have `rowid_column` is because those objects doesn't support data modification. As you can see from [the doc](https://supabase.github.io/wrappers/stripe/), `rowid_column` is needed only for data modification and those supported objects are:

- Customers
- Products
- Subscriptions

All the other objects can only support data read, so we didn't put `rowid_column` for them. Although you can still specify it, it will just be ignored in the query.

For testing locally, you need to install Docker first and then start the mock services using below command:

```
cd wrappers
docker-compose -f .ci/docker-compose.yaml -d
```

And then start `pgx`, it will start a psql client session for you:

```
cargo pgx run --features stripe_fdw
```

And then use same SQL scripts and credentials in [the test file](https://github.com/supabase/wrappers/blob/main/wrappers/src/fdw/stripe_fdw/tests.rs) to do whatever you want.

nielsbrakel:
Thanks for the feedback, I will close this PR and the other documentation changes will be included in my next PR.

Issue #86, opened 2023-05-01
Replace references to PGX with references to PGRX


dventimihasura:
# Improve documentation

## Link 

https://github.com/supabase/wrappers

## Describe the problem

The documentation refers to PGX, but evidently PGX is now PGRX.  Perhaps its name changed, but in any case, this changes the installation instructions and leads to confusion.

## Describe the improvement

Change all references to PGX, to PGRX.  For instance:

1. Built on top of pgx, providing higher level interfaces, without hiding lower-level C APIs. -> Built on top of pgrx, providing higher level interfaces, without hiding lower-level C APIs.
2. Wrappers is a pgx extension, you can follow the pgx installation steps to install Wrappers. -> Wrappers is a pgrx extension, you can follow the pgrx installation steps to install Wrappers.
3. cargo pgx install --pg-config [path_to_pg_config] --features stripe_fdw -> cargo pgrx install --pg-config [path_to_pg_config] --features stripe_fdw
4. Run it using pgx with feature -> Run it using pgrx with feature
5. cargo pgrx run pg14 --features helloworld_fdw
6. cargo pgx test --features all_fdws,pg15 -> cargo pgrx test --features all_fdws,pg15
7. Windows is not supported, that limitation inherits from pgrx -> Windows is not supported, that limitation inherits from pgrx

dventimihasura:
Nevermind.  Maybe it's just enough to point out that pgx became pgrx between 0.7.4 and 0.8.0, that it introduced breaking changes when it did so, and that therefore for Supabase wrappers you should not follow the exact instructions on the pg(r)x site and instead install pgx.  I.e.,

Instead of 

```shell
cargo install --locked cargo-pgrx
```

do

```shell
cargo install --locked cargo-pgx
```

dventimihasura:
Addendum. Evidently, wrappers requires pgx 0.6.1, not 0.7.4:

```shell
cargo install --locked --version=0.6.1 cargo-pgx
```

burmecia:
Thanks @dventimihasura , you're right, the wrappers can only support pgx 0.6.1, and we're in the progress to catch up with pgx's latest changes.

PR #99, opened 2023-05-28
upgrade pgx 0.6 to pgrx 0.8.3

tedverse:
Major dependency (pgx) update. Inspired by https://github.com/supabase/pg_graphql/pull/358. 

Currently wrappers use pgx 0.6.1, which was prior to the the major rename of pgx to pgrx.
https://github.com/supabase/wrappers/issues/86

## What is the new behavior?
The old plugins should work like before without additional modification.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -37,8 +37,8 @@ jobs:
       matrix:
         extension_name:
           - wrappers
-        pgx_version:
-          - 0.6.1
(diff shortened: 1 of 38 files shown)


burmecia:
Thank you very much @tedverse! That's a big change, let us have some tests. Also could you upgrade pgrx to latest v0.8.4?

burmecia (approved):

PR #109, opened 2023-07-05
feat: add logflare fdw

burmecia:

This PR is to add the Logflare FDW.

N/A

## What is the new behavior?

N/A

## Additional context

N/A

### DIFF
--- a/docs/logflare.md
+++ b/docs/logflare.md
@@ -1,0 +1,91 @@
+[Logflare](https://logflare.app) is a centralized web-based log management solution to easily access Cloudflare, Vercel & Elixir logs.
+
+### Wrapper 
+To get started with the Logflare wrapper, create a foreign data wrapper specifying `handler` and `validator` as below.
+
+```sql
+create extension if not exists wrappers;
+
+create foreign data wrapper logflare_wrapper
+  handler logflare_fdw_handler
+  validator logflare_fdw_validator;
+```
+
+### Server 
+
+Next, we need to create a server for the FDW to hold options and credentials.
+
+#### Auth (Supabase)
+
+If you are using the Supabase platform, this is the recommended approach for securing your [Logflare access token](https://docs.logflare.app/concepts/access-tokens/).
+
+Create a secure key using pgsodium
+```sql
+select pgsodium.create_key(name := 'logflare');
+```
+
+Save your Logflare access token in Vault and retrieve the `key_id`
+```sql
+insert into vault.secrets (secret, key_id)
+values (
+  'xxx',
+  (select id from pgsodium.valid_key where name = 'logflare')
+)
+returning
+	key_id;
+```
+
+Create the foreign server
+```sql
+create server logflare_server
+  foreign data wrapper logflare_wrapper
+  options (
+    api_key_id '<your key_id from above>'
+  );
+```
+
+#### Auth (Insecure)
+
+If the platform you are using does not support `pgsodium` and `Vault` you can create a server by storing your [Logflare Access Token](https://docs.logflare.app/concepts/access-tokens/) directly.
+
+
+!!! important
+
+    Credentials stored using this method can be viewed as plain text by anyone with access to `pg_catalog.pg_foreign_server`
+
+```sql
+create server logflare_server
+   foreign data wrapper logflare_wrapper
+   options (
+     api_key 'xxx'
+   );
+```
+
+
+### Tables
+
+Logflare wrapper is implemented with [ELT](https://hevodata.com/learn/etl-vs-elt/) approach, so the data transformation is encouraged to be performed locally after data is extracted from remote data source.
+
+
+#### Foreign Table Options
+
+The full list of foreign table options are below:
+
+- `endpoint` - Logflare endpoint, required.
+
+#### Examples
+
+```sql
+create foreign table people (
+  id bigint,
+  name text,
+  ts timestamp
+)
+  server logflare_server
+  options (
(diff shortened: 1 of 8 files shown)


wrappers/src/fdw/logflare_fdw/logflare_fdw.rs
```diff
@@ -0,0 +1,317 @@
...
+
+fn extract_params(quals: &[Qual]) -> Option<Vec<Qual>> {
+    let params = quals
+        .iter()
+        .filter_map(|q| {
+            if q.field.starts_with("_param_") {
+                Some(q.clone())
+            } else {
+                None
+            }
```
olirice, line 52: does this silently ignore components of the `where` clause that don't start with `_param_` or are the other filters applied elsewhere?

docs/logflare.md
```diff
@@ -0,0 +1,171 @@
...
+
+Save your Logflare access token in Vault and retrieve the `key_id`
+```sql
+insert into vault.secrets (secret, key_id)
+values (
+  'xxx',
+  (select id from pgsodium.valid_key where name = 'logflare')
+)
+returning
+	key_id;
```
olirice, line 35: (nit) extra indent

docs/logflare.md
```diff
@@ -0,0 +1,171 @@
...
+```sql
+create foreign table runtime_hours (
+  db_size text,
+  org_id text,
+  runtime_hours numeric,
+  runtime_minutes bigint,
+  _param_org_id bigint,
+  _param_iso_timestamp_start text,
+  _param_iso_timestamp_end text,
+  _attrs text
```
olirice, line 149: what happens if you select any of these columns in the response?

docs/logflare.md
```diff
@@ -0,0 +1,171 @@
...
+select
+  db_size,
+  org_id,
+  runtime_hours,
+  runtime_minutes
+from
+  runtime_hours
+where _param_org_id = 123
+  and _param_iso_timestamp_start = '2023-07-01 02:03:04'
+  and _param_iso_timestamp_end = '2023-07-02';
```
olirice, line 169: this feels a little unusual in the context of a FDW (vs function that returns a table). Did we brainstorm this interface on notion? If not, its there a similar interface precedent?

olirice (commented):
the implementation looks great

----

Lets wait for @chasers to get back so he can weigh in on the interface

We should also figure out how to test this locally before merging. @chasers can logflare endpoints run in a e.g. container?

docs/logflare.md
burmecia: If those meta columns are in select list, they will show the input value as is.

docs/logflare.md
burmecia: yes it is a bit odd to use sql pass in parameters, we can use function but that will make it meaningless to use Wrappers framework. Some other fdws use foreign table `options` to pass in parameter, but that cannot be dynamic.

wrappers/src/fdw/logflare_fdw/logflare_fdw.rs
burmecia: no it won't, that is just to find `_param_` conditions in `where` clause and pass them to api call, the other conditions will still apply after data is retrieved from logflare.

docs/logflare.md
```diff
@@ -0,0 +1,171 @@
...
+
+```sql
+create foreign table people (
+  id bigint,
+  name text,
+  _attrs text
+)
+  server logflare_server
+  options (
+    endpoint '9dd9a6f6-8e9b-4fa4-b682-4f2f5cd99da3'
```
Ziinc, line 111: just a small point, it can be either uuid or the endpoint name, such as `runtime_hours.prod`

docs/logflare.md
```diff
@@ -0,0 +1,171 @@
...
+
+#### Foreign Table Options
+
+The full list of foreign table options are below:
+
+- `endpoint` - Logflare endpoint, required.
+
+#### Meta Column
+
+You can define a specific meta column `_attrs` (data type: `text`) in the foreign table. It will store the whole record in JSON string format, so you can extract any fields from it using Postgres JSON queries like `_attrs::json->>'foo'`. See more examples below.
```
Ziinc, line 78: would it be better to call it `_result`, such that it corresponds to the payload returned from the LF endpoint?
such that `_result` would be the result set, then `_error` would be the api errors if any

Ziinc (approved):

Ziinc:
For integration testing, LF can run in docker container with supabase mode enabled so that the supabase-related logflare endpoints are seeded, then we can run test queries against it.

docs/logflare.md
burmecia: good point, I will change it to `_result`, api errors will be reported by standard Postgres error reporting functions, so the `_error` may not needed.

Ziinc: If it is a Bigquery error (such as syntax error) , api would return 200 with a error key.
Since we want pg to handle errors, we should panic if there is an error message returned by the api then.

docs/logflare.md
burmecia: the `endpoint` option here is just a string to be appended to the request url, so using name is absolutely fine, I will update the doc anyway.

burmecia:
> For integration testing, LF can run in docker container with supabase mode enabled so that the supabase-related logflare endpoints are seeded, then we can run test queries against it.

Does the LF container need a BigQuery backend or can we use other local backend? As the integration testing will be running in GitHub action, it might not able to access cloud-based backends.

PR #116, opened 2023-07-15
chore: minor fixes

glarrain-cdd:

Minor fixes.

- A few config files are not strictly compliant with the corresponding JSON or YAML schemas.
- The PDPG APT repo URL used for downloading the PostgreSQL packages does not use HTTPS.

## What is the new behavior?

- A few config files are now strictly compliant with the corresponding JSON or YAML schemas.
- The PDPG APT repo URL used for downloading the PostgreSQL packages uses HTTPS.

## Additional context

None

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -14,7 +14,6 @@ node_modules/
 results/
 venv/
 *.md
-*.md
 *.py
 *.yaml
 *.yml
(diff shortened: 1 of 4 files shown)


burmecia (approved):
Thanks for the PR!

PR #122, opened 2023-07-28
Fix WHERE pushdown with parametrized SQL as table clickhouse FDW

ibotty:

Bug fix

`SELECT`ing with `WHERE` clause that only contains the parameter from the parametrized SQL table definition was inserting a spurious `WHERE` without any clause. 

## What is the new behavior?

Checks for empty conditions and only adds `WHERE` when appropriate.

Also adds some tests.


burmecia (approved):

burmecia:
Thank you for the PR!

PR #123, opened 2023-08-03
Foreign data wrapper with stripe docs update

zer0eXploit:

Docs update

Cannot query stripe prices table after creating the table following the guide from supabase docs. When querying the following error is returned.

```object "pricing" not implemented.```

## What is the new behavior?

The table can now be queried and stripe price data are returned.

## Additional context

I am developing in stripe test mode with strip test api keys.


burmecia (approved):
Thanks for the PR!

PR #134, opened 2023-08-30
feat: add view support for airtable fdw

burmecia:

This PR is to add support for [Airtable View](https://support.airtable.com/docs/view-basics), which was requested in #133. 

The current Airtable fdw only supports `table`, `view` is not supported.

## What is the new behavior?

Airtable View is supported by a new foreign table option `view_id`.

## Additional context

N/A


imor (approved):
LGTM

PR #143, opened 2023-09-11
fix: make wrappers project part of workspace

imor:

A change to how the project is organized.

The `wrappers` project is not part of the workspace. This has a few drawbacks:

1. IDEs like `rust-analyzer` or IntelliJ IDEA Rust plugin don't work if the project root is opened in them. For example, I can't go to definition across projects.
2. There are separate `target` folders in each project. If two projects depend upon a single dependency, its artefacts will be copied to each `target` folder separately, increasing disc usage.
3. Cargo commands like test, fmt, clippy etc. need to be run separately in each folder increasing dev friction.

## What is the new behavior?

The `wrappers` project is part of the workspace. This fixes the above drawbacks and makes the project work with IDEs, with less disc usage and easier cargo usage during development and in GH actions.

Following changes were done as part of this:

1. File `wrappers/.cargo/config.toml` was moved to `.cargo/config.toml` as mentioned in the [`pgrx` docs](https://github.com/pgcentralfoundation/pgrx/blob/develop/cargo-pgrx/README.md#creating-a-new-extension) about moving to a workspace.
4. All three subprojects, `wrappers`, `supabase-wrappers` and `supabase-wrappers-macros` are part of the workspace in the top level `Cargo.toml`. In addition, the top level `Cargo.toml` contains the profiles now, fixing the [profiles warning](https://github.com/supabase/wrappers/pull/143#issuecomment-1728834552).
5. Set `"rust-analyzer.cargo.features": ["all_fdws,helloworld_fdw"]` in `.vscode/settings.json` to make `rust-analyzer` in VSCode recognize all rust files in `wrappers/src/fdw` folder as part of the workspace.
6. In order to avoid accidentally publishing `wrappers` to crates.io, a `publish = false` line is added to its `Cargo.toml` file.
7. Update release workflow to look for build artefacts in workspace's `target` directory rather than in `wrappers/target`.

## Additional Context

~I also had to pin the `clickhouse-rs` to a specific commit in the `Cargo.toml` depdencies because of its accidental upgrade when adding `wrappers` to the workspace.~ Edit: `clickhouse-rs` dependency is already pinned to a specific revision in the main branch which is merged in this PR's branch.


imor:
When member crates have profiles building the workspace emits a warning:

```
warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
```

There's an open issue in cargo for this [here](https://github.com/rust-lang/cargo/issues/8264). We need to find a way around this before this can be merged.

burmecia (approved):

Issue #81, opened 2023-04-07
Return `Result<T, E>` from trait methods


Jesse-Bakker:
Currently, error handling is implemented the way it is in pgx, which is by `panic`ing and catching the panic at the transaction boundary. This is good at the interface with postgres, but for a higher level API, using the standard rust `Result` method of error handling is generally more ergonomic. This because interfaces signal fallibility and it allows using the standard machinery for handling and converting errors.

```rust
struct HelloWorldFdw{ url: String }

impl ForeignDataWrapper for HelloWorldFdw {
    fn new(options: &HashMap<String, String>) -> Self {
        Self {
            url: require_option("url", options).unwrap()
        }
    }
}
```

Here, `fn new` is fallible, but nothing signals its fallibility to the casual reader or implementer.
`require_option` diverges, so the `unwrap` is safe (actually require_option could return `String` instead of `Option<String>`, because if it would be `None`, it would panic through `report_error`).
 If `new` returned `Result<Self, WrappersError>`, this could be rewritten as, for example:

```rust
struct HelloWorldFdw{ url: String }

impl ForeignDataWrapper for HelloWorldFdw {
    fn new(options: &HashMap<String, String>) -> Result<Self, WrappersError> {
        Ok(Self {
            url: require_option("url", options)?
        })
    }
}
```

Where `require_option()` returns a `Result<String, WrappersError>`, with the error being a `WrappersError::OptionNameNotFound(String)`, with the string representing the option name. It is now clear that `fn new` is fallible, and `require_option` is actually just `options.get(<name>).ok_or(WrappersError::OptionNameNotFound(<name>))`
Of course other errors could also be returned from `new()`. I imagine `WrappersError` to look something like:

```rust
enum WrappersError {
    OptionNameNotFound(String),
    ColumnNameNotFound(String),
    InvalidDataType(String),
    // ... etc
    Other(Box<dyn Error>),
}

impl WrappersError {
    fn error_code(&self) -> PgSqlErrorCode {
        match self {
            OptionNameNotFound(_) => PgSqlErrorCode::ERROR_CODE_FDW_OPTION_NAME_NOT_FOUND
        }
    }
}

impl Display for WrappersError{ /*...*/}
```

In `FdwState`, errors would be converted to the panic-y kind:

```rust
match Wrapper::new(options) {
    Err(e) => report_error(e.error_code(), &e.to_string()),
    Ok(wrapper) => wrapper,
}
```

I'd be happy to make a PR implementing this by the way.

[...]

PR #144, opened 2023-09-12
Fallible ForeignDataWrapper trait methods

imor:

Refactoring

The `ForeignDataWrapper` trait's methods do not return `Result`s even though they are fallible. This results in forcing the user to call `unwrap` or `expect`, instead of the more ergonomic alternative of returning an `Err`.

## What is the new behavior?

The `ForeignDataWrapper` trait's methods will now return a `Result`. The error types should be convertible into an `ErrorReport`, i.e. they should `impl From<ErrorType> for ErrorReport` for the FDW's `ErrorType`. An `ErrorReport` is a `pgrx` type to represent errors which can be reported to Postgres. The errors returned by the trait's methods will be automatically reported to Postgres by the `supabase-wrappers` framework.

An example FDW with the new API looks something like this:

```rust
/// The error type has to be specified in the `error_type` field of the proc macro
#[wrappers_fdw(
    version = "0.1.0",
    author = "Supabase",
    website = "https://github.com/supabase/wrappers/tree/main/wrappers/src/fdw/helloworld_fdw",
    error_type = "HelloWorldFdwError"
)]
pub(crate) struct HelloWorldFdw {}

/// A type representing the errors returned from the ForeignDataWrapper trait's methods
enum HelloWorldFdwError {
    InvalidFoo,
    MissingBar,
}

/// A required impl for the error type to convert into ErrorReport
impl From<HelloWorldFdwError> for ErrorReport {
    fn from(value: HelloWorldFdwError) -> Self {
        let error_message = match value {
            HelloWorldFdwError::InvalidFoo => "foo is invalid",
            HelloWorldFdwError::MissingBar => "bar is missing",
        };
        // The error code and error message to return to Postgres
        ErrorReport::new(PgSqlErrorCode::ERRCODE_FDW_ERROR, error_message, "")
    }
}

impl ForeignDataWrapper<HelloWorldFdwError> for HelloWorldFdwError {
    fn new(options: &HashMap<String, String>) -> Result<Self, HelloWorldFdwError>
    where
        Self: Sized,
    {
        // Return errors from your methods, these will be automatically reported by `supabase-wrappers`
        Err(HelloWorldFdwError::InvalidFoo)
    }

    fn begin_scan(
        &mut self,
        _quals: &[Qual],
        _columns: &[Column],
        _sorts: &[Sort],
        _limit: &Option<Limit>,
        _options: &HashMap<String, String>,
    ) -> Result<(), HelloWorldFdwError> {
        // Return another error
        Err(HelloWorldFdwError::MissingBar)
    }

    fn iter_scan(&mut self, row: &mut Row) -> Result<Option<()>, HelloWorldFdwError> {
        // Return ok if no error
        Ok(None)
    }

    fn end_scan(&mut self) -> Result<(), HelloWorldFdwError> {
        // Return ok if no error
        Ok(())
    }
}

```

## Additional context

This impl is similar to but not exactly the same as suggested in https://github.com/supabase/wrappers/issues/81. The major difference is the there is no `WrappersError` but instead each FDW can define it's own error type. A type similar to `WrappersError` can be introduced later if needed.

This PR just lays the groundwork for all FDWs to better report and handle errors. But the FDWs haven't been updated with this PR. They will be done in separate, later PRs.

### DIFF
--- a/supabase-wrappers-macros/src/lib.rs
+++ b/supabase-wrappers-macros/src/lib.rs
@@ -56,14 +56,14 @@ pub fn wrappers_fdw(attr: TokenStream, item: TokenStream) -> TokenStream {
         }
     }
 
-    if error_type.is_none() {
+    let error_type_ident = if let Some(error_type) = error_type {
+        format_ident!("{}", error_type)
+    } else {
         let quoted = quote! {
             compile_error!("Missing `error_type` in the `wrappers_fdw` attribute");
         };
         return quoted.into();
-    }
-
-    let error_type_ident = format_ident!("{}", error_type.unwrap());
+    };
 
     let item: ItemStruct = parse_macro_input!(item as ItemStruct);
     let item_tokens = item.to_token_stream();
--- a/supabase-wrappers-macros/src/lib.rs
+++ b/supabase-wrappers-macros/src/lib.rs
@@ -41,15 +41,29 @@ pub fn wrappers_fdw(attr: TokenStream, item: TokenStream) -> TokenStream {
     let mut metas = TokenStream2::new();
     let meta_attrs: Punctuated<MetaNameValue, Token![,]> =
         parse_macro_input!(attr with Punctuated::parse_terminated);
+    let mut error_type: Option<String> = None;
     for attr in meta_attrs {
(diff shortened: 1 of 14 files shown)


burmecia (approved):
Thanks for the PR!

PR #155, opened 2023-09-20
feat: return Row from iter_scan instead of mutating in-place

imor:
This PR changes the `iter_scan` method signature from `fn iter_scan(&mut self, row: &mut Row) -> Result<Option<()>, E>` to `fn iter_scan(&mut self) -> Result<Option<Row>, E>`. The new signature is better in that the user is forced to return a `Row` instead of updating the incoming row in-place, which they can forget.

### DIFF
--- a/supabase-wrappers/src/interface.rs
+++ b/supabase-wrappers/src/interface.rs
@@ -13,7 +13,6 @@ use std::collections::HashMap;
 use std::ffi::CStr;
 use std::fmt;
 use std::iter::Zip;
-use std::mem;
 use std::slice::Iter;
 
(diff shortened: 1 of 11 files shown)


burmecia:
The original design was to return a `Row` in `iter_scan`, but there was a performance issue with that design (see details #46), so we changed to use a borrowed `Row` in #48. IMO, we shouldn't cover this user negligence as it is the basic interface requirement, and can be easily identified.

imor:
Oh I see, reusing the row saves some allocations because the `Vec`s in the row don't need to grow again for each new row. Closing for now.

PR #156, opened 2023-09-20
feat: add support for qdrant foreign data wrapper

imor:
This PR adds a foreign data wrapper for [qdrant](https://qdrant.tech/) vector database. This is a bare-bones version only supporting simple select statements which returns data from the Qdrant's [scroll points](https://qdrant.github.io/qdrant/redoc/index.html#tag/points/operation/scroll_points) endpoint. For more information about how the wrapper works, read the docs in the PR.

Implementation wise, this wrapper is quite similar to others in design in that we call the `collection/<collection_name>/points/scroll` api with appropriate parameters in the wrapper callbacks and return that data in a form Postgres is happy with. One slight improvement is that we run the query in batches so as to not balloon the memory used by the wrapper. The default batch size is currently hardcoded to 1000.


olirice (commented):
*On Hold until Oct 14

PR #173, opened 2023-10-12
chore: bump version to 0.1.17

imor:
Bump version to prepare for release.


kamilogorek:
Should https://github.com/supabase/wrappers/blob/main/supabase-wrappers-macros/Cargo.toml be also updated to keep versions in sync?

imor:
> Should https://github.com/supabase/wrappers/blob/main/supabase-wrappers-macros/Cargo.toml be also updated to keep versions in sync?

No, `supabase-wrappers-macros` and `supabase-wrappers` are the "framework" crates which the `wrappers` crate depends on. Their version should be bumped only when they are released as a crate (which we should do sometime soon since we have had breaking API changes in them recently). It is ok to not update because we have a path based dependency on them, instead of a version dependency: https://github.com/supabase/wrappers/blob/f256e95843a934fe75633a16adbc357f97ec95dd/wrappers/Cargo.toml#L39

Having said that, I think it is a good idea to depend on a specific version of the `supabase-wrappers` crate because it allows us to dogfood the crate better. I have created an issue to track it [here](https://github.com/supabase/wrappers/issues/174). 

Thanks for the feedback @kamilogorek 🙏.

olirice (approved):

PR #181, opened 2023-10-18
feat: Add support for jsonb data type to Airtable

kamilogorek:
- feat: Add support for jsonb data type to Airtable
- build: Remove unused pg11-13 versions
- test: Include Airtable tests in the run
- misc: Fix docs code snippet


imor (approved):
LGTM, thanks.

wrappers/src/fdw/airtable_fdw/result.rs
```diff
@@ -212,7 +213,19 @@ impl AirtableRecord {
                         }
                     },
                 ),
-                _ => return Err(AirtableFdwError::UnsupportedColumnType(col.name.clone())),
+                pg_sys::JSONBOID => self.fields.0.get(&col.name).map_or_else(
+                    || Ok(None),
+                    |val| {
+                        if val.is_array() || val.is_object() {
```
imor, line 219: Since this is a json object, any kind of value is ok to pass in pgrx::JsonB. Any specific reason why only arrays and objects are allowed here?

kamilogorek: We can make it accept anything, however my thinking was that if someone uses `jsonb` for numeric or string types, it's not something that they were intending, and they should be using `text` or any other primitive data type.

kamilogorek: In the future, we can add support for `TEXTARRAYOID` and all its siblings, which would allow using `text[]` type, but I havent seen anyone requesting support for it so far.

imor: Okay, let's go with the current implementation.

PR #188, opened 2023-11-02
fix: only update `wrappers_fdw_stats` table in a read-write transaction

imor:

Bug fix

If a query on a foreign table was done in a read-only transaction, it failed with this error:

```
ERROR:  cannot execute INSERT in a read-only transaction
CONTEXT:  SQL statement "insert into public.wrappers_fdw_stats as s (fdw_name, create_times) values($1, $2)
         on conflict(fdw_name)
         do update set
            create_times = coalesce(s.create_times, 0) + excluded.create_times,
            updated_at = timezone('utc'::text, now())"
```

## What is the new behavior?

This error occurred because `wrappers` updated `wrappers_fdw_stats` table for each query on a foreign table. `wrappers` will now update the `wrappers_fdw_stats` table only if the transaction is read-write.

## Additional context

N/A

### DIFF
--- a/wrappers/src/stats.rs
+++ b/wrappers/src/stats.rs
@@ -43,26 +43,35 @@ fn get_stats_table() -> String {
 // increase stats value
 #[allow(dead_code)]
 pub(crate) fn inc_stats(fdw_name: &str, metric: Metric, inc: i64) {
-    let sql = format!(
-        "insert into {} as s (fdw_name, {}) values($1, $2)
+    if !is_txn_read_only() {
+        let sql = format!(
+            "insert into {} as s (fdw_name, {}) values($1, $2)
          on conflict(fdw_name)
          do update set
             {} = coalesce(s.{}, 0) + excluded.{},
             updated_at = timezone('utc'::text, now())",
-        get_stats_table(),
-        metric,
-        metric,
-        metric,
-        metric
-    );
-    Spi::run_with_args(
-        &sql,
-        Some(vec![
-            (PgBuiltInOids::TEXTOID.oid(), fdw_name.into_datum()),
-            (PgBuiltInOids::INT8OID.oid(), inc.into_datum()),
-        ]),
-    )
-    .unwrap();
+            get_stats_table(),
+            metric,
+            metric,
+            metric,
+            metric
+        );
+        Spi::run_with_args(
+            &sql,
+            Some(vec![
+                (PgBuiltInOids::TEXTOID.oid(), fdw_name.into_datum()),
+                (PgBuiltInOids::INT8OID.oid(), inc.into_datum()),
+            ]),
+        )
+        .unwrap();
+    }
+}
+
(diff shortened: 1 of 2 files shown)


wrappers/src/stats.rs
```diff
@@ -43,26 +43,35 @@ fn get_stats_table() -> String {
 // increase stats value
 #[allow(dead_code)]
 pub(crate) fn inc_stats(fdw_name: &str, metric: Metric, inc: i64) {
-    let sql = format!(
-        "insert into {} as s (fdw_name, {}) values($1, $2)
+    if !is_txn_read_only() {
+        let sql = format!(
+            "insert into {} as s (fdw_name, {}) values($1, $2)
          on conflict(fdw_name)
```
burmecia, line 49: can we add indent here to line up the sql?

wrappers/src/stats.rs
```diff
@@ -43,26 +43,35 @@ fn get_stats_table() -> String {
...
+            Some(vec![
+                (PgBuiltInOids::TEXTOID.oid(), fdw_name.into_datum()),
+                (PgBuiltInOids::INT8OID.oid(), inc.into_datum()),
+            ]),
+        )
+        .unwrap();
+    }
+}
+
+fn is_txn_read_only() -> bool {
```
burmecia, line 70: this could be simplified to:
```
#[inline]
fn is_txn_read_only() -> bool {
    Spi::get_one("show transaction_read_only") == Ok(Some("on"))
}
```

wrappers/src/stats.rs
```diff
@@ -43,26 +43,35 @@ fn get_stats_table() -> String {
 // increase stats value
 #[allow(dead_code)]
 pub(crate) fn inc_stats(fdw_name: &str, metric: Metric, inc: i64) {
-    let sql = format!(
-        "insert into {} as s (fdw_name, {}) values($1, $2)
+    if !is_txn_read_only() {
```
burmecia, line 46: can we use early exit here?
```
if is_txn_read_only() {
    return;
}

...<rest code>...
```
Also, can we move the `is_txn_read_only` function def to top of the page?

wrappers/src/stats.rs
imor: Nice!

wrappers/src/stats.rs
imor: Done.

wrappers/src/stats.rs
imor: Done.

burmecia (approved):

StanGirard:
@imor issue still occurs on hosted version with latest version of stripe with a fresh new project.

imor:
@StanGirard we haven't yet rolled out an image which includes the fixed version of `wrappers` (hope to do so in a couple of days). If it's urgent for you we can bump `wrappers` version for your project if you can raise a support ticket.

PR #191, opened 2023-11-03
add cnosdb wrapper

Subsegment:

Feature

https://github.com/supabase/wrappers/discussions/187

## What is the new behavior?

Add a new fdw for cnosdb


Subsegment:
What do you think of this? @burmecia

burmecia:
Thanks for the PR and sorry for the very late reply. This work is impressive but unfortunately we don't see much CnosDB use cases or requests on Supabase platform, so we'll just close this PR for now as it is there for so long. If you do really want to use it with Supabase, [self-hosting](https://supabase.com/docs/guides/self-hosting) might be an option. Another option is [Wasm FDW](https://github.com/supabase-community/wasm-fdw-example), it might be possible to build CnosDB FDW using Wasm and then you can use it on Supabase platform.

PR #192, opened 2023-11-13
fix: incorrect stripe table join

### LINKED ISSUE #113: Stripe joins failing to return results.
# Bug report

## Describe the bug

Joining stripe fdw tables seems to fail or return incomplete results. I'm not sure if this is expected and just not yet implemented or a bug or what so opening an issue before I dig deeper.

e.g.
```sql
SELECT stripe_customers.email, stripe_subscriptions.currency 
FROM stripe_customers 
INNER JOIN stripe_subscriptions ON stripe_customers.id = stripe_subscriptions.customer
```
Using the mock data returns nothing and using our stripe account data returns a single row even though there are additional valid results.

## To Reproduce

Add the following test and run the test suite.

```rust
  let results = c
                .select(
                    "SELECT stripe_customers.email, stripe_subscriptions.currency FROM stripe_customers INNER JOIN stripe_subscriptions ON stripe_customers.id = stripe_subscriptions.customer",
                    None,
                    None,
                ).unwrap()
                .filter_map(|r| {
                    r.get_by_name::<&str, _>("email")
                        .unwrap()
                        .zip(r.get_by_name::<&str, _>("currency").unwrap())
                })
                .collect::<Vec<_>>();
            assert_ne!(results, vec![]);
```
Run the query above against a stripe account with customers and subscriptions.

## Expected behavior

Joined rows are returned.

burmecia:

This PR is to fix #113 . It will keep remote scan result and add implementation for `re_scan()` callback to reset iteration index when in nested loop join.

If the query joins 2 Stripe foreign tables, the result data isn't correct.

## What is the new behavior?

Stripe foreign table join should work correctly.

## Additional context

We may need to implement the same `re_scan()` callback for other FDWs to support join on foreign tables, but this PR only deals with Stripe FDW for now.

### DIFF
--- a/docs/stripe.md
+++ b/docs/stripe.md
@@ -766,10 +766,6 @@ While any column is allowed in a where clause, it is most efficient to filter by
 - id
 - destination
 
-## Limitations
-
-Joining local data against a Stripe FDW returns incomplete or empty results. We're tracking the [issue on GitHub](https://github.com/supabase/wrappers/issues/113) and working towards a fix.
-
 ## Examples
 
 Some examples on how to use Stripe foreign tables.
--- a/wrappers/src/fdw/stripe_fdw/stripe_fdw.rs
+++ b/wrappers/src/fdw/stripe_fdw/stripe_fdw.rs
@@ -261,6 +261,7 @@ pub(crate) struct StripeFdw {
     scan_result: Option<Vec<Row>>,
     obj: String,
     rowid_col: String,
+    iter_idx: usize,
 }
 
 impl StripeFdw {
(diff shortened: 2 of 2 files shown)


imor (approved):

napter:
@burmecia this is still not working in the supabase sql editor.

This returns 1 result:
```select * from stripe.customers c where c.email = 'xxx'```

This returns 5 results:
select * from stripe.payment_intents where customer = 'xxx'

However, this query returns only 1 result instead of 5 and payment intents is always null
```select * from stripe.customers c left join stripe.payment_intents p on (p.customer = c.id) where c.email = 'xxxx'```

**Comment by burmecia** [comment]
can you try use `customer.id` as the filter instead of email?

```sql
select * from stripe.customers c left join stripe.payment_intents p on (p.customer = c.id) where c.id = 'cus_xxxx'
```

PR #183, opened 2023-10-26
feat: add auth0 wrapper

J0:
## What kind of change does this PR introduce?

POC around an auth0 wrapper for continuous migration from Auth0 to Supabase Auth. Done as a side project.

We currently support only the `users` page. Similar to the Stripe FDW more objects can be added via the `options` table. For the moment, we only support the key fields from users (e.g. identities, email, email_verified) though more can be added upon request.

CleanShot 2023-11-22 at 10 22 02@2x

Not too familar with Rust so will ask in advance for the patience of reviewers when reviewing this PR 🙏 

- [x] Mock server + tests
- [x] Refactor to align with `qdrant_fdw` structure

PR #206, opened 2023-12-10
feat: Auth0 add pagination

J0:
FLUP to #183 where we aim to introduce pagination on the user object. Auth0 has `pagination` by supplying the following fields:

- `total` - this is the total number of entries
-  `per_page` - this is the number of entries per page
- `page` - this is the current page we're on

Here are the [docs](https://auth0.com/docs/manage-users/user-search/view-search-results-by-page#sample-request)

We default to having 50 entries per_page. It also seems like there's a 1000 entry limit which is unfortunate.

No proper pagination

### DIFF
--- a/wrappers/src/fdw/auth0_fdw/auth0_client/rows_iterator.rs
+++ b/wrappers/src/fdw/auth0_fdw/auth0_client/rows_iterator.rs
@@ -42,7 +42,10 @@ impl RowsIterator {
             .map(|u| u.into_row(&self.columns))
             .collect();
         self.page_offset = self.page_offset.map(|offset| offset + 1);
-        let page_offset = self.page_offset.unwrap_or(0);
+        let page_offset = match self.page_offset {
+            Some(offset) => offset,
+            None => return Err(Auth0ClientError::MissingPageOffset),
+        };
         self.have_more_rows = total > self.per_page * page_offset;
         Ok(self.get_next_row())
     }
--- a/wrappers/src/fdw/auth0_fdw/auth0_client/rows_iterator.rs
+++ b/wrappers/src/fdw/auth0_fdw/auth0_client/rows_iterator.rs
@@ -8,7 +8,7 @@ pub(crate) struct RowsIterator {
     columns: Vec<Column>,
     rows: VecDeque<Row>,
     have_more_rows: bool,
-    page_offset: Option<u64>,
+    page_offset: u64,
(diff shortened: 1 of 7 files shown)


J0:
Coming back to this

wrappers/src/fdw/auth0_fdw/auth0_client/rows_iterator.rs
```diff
@@ -4,46 +4,49 @@ use supabase_wrappers::prelude::{Column, Row};
...
-        // self.next_page_offset = user_result.next_page_offset;
-        // self.have_more_rows = self.next_page_offset.is_some();
-        //TODO: add proper logic to figure out when to set have_more_rows to false
-        //hardcoding to false just to make the tests pass for now
-        self.have_more_rows = false;
+        self.page_offset = self.page_offset.map(|offset| offset + 1);
+        let page_offset = match self.page_offset {
+            Some(offset) => offset,
+            None => return Err(Auth0ClientError::MissingPageOffset),
+        };
```
imor, line 48: If `page_offset`'s type is changed from an `Option<u64>` to `u64` the code simplifies to:

```suggestion
        self.page_offset += 1;
```

`Option<u64>` is not needed because `page_offset` is initialized to `Some(0)` and never set to `None`. `Auth0ClientError::MissingPageOffset` variant can also be removed then.

imor:
> It also seems like there's a 1000 entry limit which is unfortunate.

This can be a serious limitation for people who have more than 1000 users in Auth0. Is there an alternative API which allows access to all users?

J0:
@imor thanks for the review and sflr. To my knowledge there are two other

> It also seems like there's a 1000 entry limit which is unfortunate.
> This can be a serious limitation for people who have more than 1000 users in Auth0. Is there an alternative API which allows access to all users?

I know of [the get users by email](https://auth0.com/docs/manage-users/user-search/retrieve-users-with-get-users-by-email-endpoint) and [get users by id](https://auth0.com/docs/manage-users/user-search/retrieve-users-with-get-users-by-id-endpoint) which are recommended. But I guess the concern there is that it'd be harder to bulk select since one would need to know the email and/or id in advance. 

Let me also poll the team

burmecia:
@J0 can you also add doc `/docs/auth0.md` for this wrapper?

burmecia (approved):

PR #207, opened 2023-12-11
fix: update with pgrx upstream and remove compilation warning

burmecia:

After merged PR #205, there are some compilation warnings in `supabase-wrappers`, we need to deal with them.

N/A

## What is the new behavior?

N/A

## Additional context

When upstream pgrx merged [this PR](https://github.com/pgcentralfoundation/pgrx/pull/1432), we need to remove the temporary warning stopper.


imor (approved):

J0:
Thanks! Sorry for the trouble

PR #215, opened 2024-01-03
feat: add redis foreign data wrapper

burmecia:

This PR is to add a new Redis foreign data wrapper, which supports basic data read at this moment.

N/A

## What is the new behavior?

N/A

## Additional context

N/A

### DIFF
--- a/docs/redis.md
+++ b/docs/redis.md
@@ -1,0 +1,326 @@
+[Redis](https://redis.io/) is an open-source in-memory storage, used as a distributed, in-memory key–value database, cache and message broker, with optional durability.
+
+The Redis Wrapper allows you to read data from Redis within your Postgres database.
+
+## Supported Redis Data Types
+
+All Redis value will be stored as `text` or `jsonb` column in Postgres, below are the supported Redis data types:
+
+| Redis Type          | Foreign Table Type (src_type)   |
+| ------------------- | ------------------------------- |
+| List                | list                            |
+| Set                 | set                             |
+| Hash                | hash                            |
+| Sorted Set          | zset                            |
+| Stream              | stream                          |
+| Multiple List       | multi_list                      |
+| Multiple Set        | multi_set                       |
+| Multiple Hash       | multi_hash                      |
+| Multiple Sorted Set | multi_zset                      |
+
+**See below for more descriptions for the `Multiple *` types and `src_type` foreign table option.**
+
+## Preparation
+
+Before you get started, make sure the `wrappers` extension is installed on your database:
+
+```sql
+create extension if not exists wrappers with schema extensions;
+```
+
+and then create the foreign data wrapper:
+
+```sql
+create foreign data wrapper redis_wrapper
+  handler redis_fdw_handler
+  validator redis_fdw_validator;
+```
+
+### Secure your credentials (optional)
+
+By default, Postgres stores FDW credentials inside `pg_catalog.pg_foreign_server` in plain text. Anyone with access to this table will be able to view these credentials. Wrappers is designed to work with [Vault](https://supabase.com/docs/guides/database/vault), which provides an additional level of security for storing credentials. We recommend using Vault to store your credentials.
+
+```sql
+-- Save your Redis connection URL in Vault and retrieve the `key_id`
+insert into vault.secrets (name, secret)
+values (
+  'redis',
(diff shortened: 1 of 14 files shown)


docs/redis.md
```diff
@@ -0,0 +1,326 @@
...
+
+### Secure your credentials (optional)
+
+By default, Postgres stores FDW credentials inside `pg_catalog.pg_foreign_server` in plain text. Anyone with access to this table will be able to view these credentials. Wrappers is designed to work with [Vault](https://supabase.com/docs/guides/database/vault), which provides an additional level of security for storing credentials. We recommend using Vault to store your credentials.
+
+```sql
+-- Save your Redis connection URL in Vault and retrieve the `key_id`
+insert into vault.secrets (name, secret)
+values (
+  'redis',
```
imor, line 47: Name the key `redis_conn_str` or something similar?

docs/redis.md
```diff
@@ -0,0 +1,326 @@
...
+  create foreign table redis_list (
+    element text
+  )
+  server redis_server
+  options (
+    src_type 'list',
+    src_key 'list'
+  );
+
+  select * from redis_list;
```
imor, line 194: Can we show the data returned by the `select` statements here just to tie in with the data added by redis cli earlier.

docs/redis.md
burmecia: sure, use `redis_conn_url`.

wrappers/src/fdw/redis_fdw/mod.rs
```diff
@@ -0,0 +1,45 @@
...
+use thiserror::Error;
+
+use supabase_wrappers::prelude::OptionsError;
+
+#[derive(Error, Debug)]
+enum RedisFdwError {
+    #[error("'{0}' source type is not supported")]
+    UnsupportedSourceType(String),
+
+    #[error("'{0}' foreign table can only has one column")]
```
imor, line 17: typo `has` -> `have`

wrappers/src/fdw/redis_fdw/mod.rs
```diff
@@ -0,0 +1,45 @@
...
+    OnlyOneColumn(String),
+
+    #[error("'{0}' foreign table can only has two column")]
```
imor, line 20: typo `has` -> `have`

wrappers/src/fdw/redis_fdw/redis_fdw.rs
```diff
@@ -0,0 +1,381 @@
...
+                stats::inc_stats(Self::FDW_NAME, stats::Metric::RowsIn, 1_i64);
+                stats::inc_stats(
+                    Self::FDW_NAME,
+                    stats::Metric::RowsOut,
+                    self.scan_result.len() as i64,
+                );
+            }
+
+            let mut tgt_row = Row::new();
+            if let Some(val) = self.scan_result.drain(0..1).last() {
```
imor, line 101: `drain`ing from the front of a `Vec` could be expensive because all the remaining items need to be copied one position down, maybe use `VecDeque` and use its `pop_front` method?

wrappers/src/fdw/redis_fdw/redis_fdw.rs
```diff
@@ -0,0 +1,381 @@
...
+                    stats::Metric::RowsOut,
+                    self.scan_result_stream.len() as i64,
+                );
+            }
+
+            let mut tgt_row = Row::new();
+            if let Some(ent) = self.scan_result_stream.drain(0..1).last() {
+                let id = ent.keys().nth(0).unwrap();
+                // Redis won't allow null value, so we can safely unwrap here
+                let items = json!(ent.values().nth(0).unwrap());
```
imor, line 171: Use `expect(msg)` instead of unwrap? The reason is that if this unwrap ever fails (if the assumption ever becomes wrong) we'll get a hint about the failure because of the `msg` passed to `expect` and that should help in debugging.

wrappers/src/fdw/redis_fdw/redis_fdw.rs
burmecia: the `redis` lib doesn't support `VecDeque` deserialization, so I will use `self.iter_idx` as the iteration index in a batch.

imor (approved):

PR #216, opened 2024-01-03
feat: upgrade S3 FDW dependencies

J0:

There is an upgrade for S3 dependencies released which replaces some of the deprecated functions. Additionally, there are some shared dependencies between the [Cognito Identity Crate](https://crates.io/crates/aws-sdk-cognitoidentity) and the S3 FDW dependencies so upgrading the dependencies would simplify the implementation fo the Cognito wrapper

We are currently on v0

## What is the new behavior?

We move to v1 of the crates

### DIFF
--- a/wrappers/src/fdw/s3_fdw/parquet.rs
+++ b/wrappers/src/fdw/s3_fdw/parquet.rs
@@ -63,7 +63,7 @@ impl S3ParquetReader {
         )
         .map_err(to_io_error)
         .map(|output| {
-            let object_size = output.object_size() as u64;
+            let object_size = output.object_size().unwrap_or_default() as u64;
             self.object_size = Some(object_size);
             object_size
         })
--- a/wrappers/src/fdw/s3_fdw/parquet.rs
+++ b/wrappers/src/fdw/s3_fdw/parquet.rs
@@ -62,10 +62,13 @@ impl S3ParquetReader {
                 .send(),
         )
         .map_err(to_io_error)
-        .map(|output| {
-            let object_size = output.object_size().unwrap_or_default() as u64;
+        .and_then(|output| {
+            let object_size = output.object_size().ok_or::<IoError>(IoError::new(
+                ErrorKind::Other,
+                "object has no size attribute",
+            ))? as u64;
             self.object_size = Some(object_size);
-            object_size
+            Ok(object_size)
         })
     }
 }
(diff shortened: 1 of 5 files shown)


wrappers/src/fdw/s3_fdw/s3_fdw.rs
```diff
@@ -170,7 +171,9 @@ impl ForeignDataWrapper<S3FdwError> for S3Fdw {
         env::set_var("AWS_ACCESS_KEY_ID", creds.0);
         env::set_var("AWS_SECRET_ACCESS_KEY", creds.1);
         env::set_var("AWS_REGION", region);
-        let config = ret.rt.block_on(aws_config::load_from_env());
+        let config = ret
+            .rt
+            .block_on(aws_config::load_defaults(BehaviorVersion::v2023_11_09()));
```
J0, line 176: Looks like there was a change in how defaults are handled. It seems like we can [opt between latest or the behaviour as of late 2023](https://docs.rs/aws-config/1.1.1/aws_config/struct.BehaviorVersion.html#method.v2023_11_09)

I've opted to use 2023 for now.

wrappers/src/fdw/s3_fdw/parquet.rs
```diff
@@ -63,7 +63,7 @@ impl S3ParquetReader {
         )
         .map_err(to_io_error)
         .map(|output| {
-            let object_size = output.object_size() as u64;
+            let object_size = output.object_size().unwrap_or_default() as u64;
```
imor, line 66: if `output.object_size()` returns `None`, `object_size` will be zero. Is that ok? I'm not too familiar with this code, maybe @burmecia can comment.

burmecia: we're assuming `object_size` always available even it is zero, so should return an error here. I've made the change.

wrappers/src/fdw/s3_fdw/s3_fdw.rs
burmecia: I think use latest behavior is fine, I've made the change.

burmecia (approved):

wrappers/src/fdw/s3_fdw/parquet.rs
J0: Thanks!

PR #221, opened 2024-01-10
docs(stripe): add docs for `api_url` server option

### LINKED ISSUE #219: Add Stripe API URL configuration and documentation
# Improve documentation

## Link 

https://supabase.github.io/wrappers/stripe/#connecting-to-stripe

## Describe the problem

Its missing, and confusing. Is there a reason why `https://api.stripe.com`  is not inserted by default?

## Describe the improvement

add that `https://api.stripe.com` is the default base URL.
Documented here`https://stripe.com/docs/api`

Without this, all DB operations fail, so making it undefined if it is not valid seems off.

burmecia:

This PR is to add docs for the missing `api_url` Stripe FDW server option. Fix #219 

N/A

## What is the new behavior?

N/A

## Additional context

N/A

### DIFF
--- a/docs/stripe.md
+++ b/docs/stripe.md
@@ -40,7 +40,8 @@ We need to provide Postgres with the credentials to connect to Stripe, and any a
     create server stripe_server
       foreign data wrapper stripe_wrapper
       options (
-        api_key_id '<key_ID>' -- The Key ID from above.
+        api_key_id '<key_ID>', -- The Key ID from above, required.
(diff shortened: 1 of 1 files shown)


imor (approved):

kyeshmz:
Is this really optional though? You cannot remove the API URL in the configuration without having to specify a stripe secret key, which is against what you are trying to do when you are doing something with vault.

Screenshot 2024-01-11 at 11 25 55 AM

PR #239, opened 2024-02-16
fix: reset state when begin scan

### LINKED ISSUE #237: Stripe FDW randomly returns no data
I am querying stripe FDW tables using an rpc call using supabase-js from my web app. About half the time, the query returns no results. It should return 2 rows, and does so about half the time. I can reproduce it reliably by just refreshing the page. It is not obvious when it will return results and when not. It may be a function of how quickly I refresh, but isn't always. I can refresh after waiting several minutes, get no results, refresh again and get results as well as vice-versa. There isn't any rhyme or reason to it as far as I can tell. When it returns no results, it does not have an error, just an empty result: `{data: [], error: null}`.

Supabase support is telling me to use pg_cron and to build what is effectively a materialized view of the stripe FDW data within supabase. If that is the recommended workflow, I think it should be documented as such. 

Here is the rpc call with the query, it joins the product and price tables. There are two products, and those were populated once a while ago and have not changed.

```
 create or replace function public.get_subscription_prices()
returns table (
    name text,
    product_id text,
    price_id text, 
    price real
)
language plpgsql security definer as
$$
#variable_conflict use_column
begin
    set statement_timeout to 60000;
    return QUERY
        select p.name, p.id as product_id, q.id as price_id, (q.unit_amount / 100.0)::real as price
        from stripe.products p
        left join stripe.prices q on p.id = q.product
        where p.active and q.active;
end
$$;
```

burmecia:

This PR is to add state reset code when begin a scan, this can partially fix #237.

The foreign table scan execution plan will be cached when foreign table query is defined in a function, this will result in `begin_scan()` being called without calling `new()` function, which is supposed to re-initialise the FDW instance.

## What is the new behavior?

Added some state reseting codes in `begin_scan()` for some FDWs, so that their inner state will be reset correctly.

## Additional context

This PR can partially fix #237, so the continuous  foreign table function call should be correct. But for permanent fix, we need to find out better way to manage and re-use `FdwState`.

### DIFF
--- a/.github/workflows/test_wrappers.yml
+++ b/.github/workflows/test_wrappers.yml
@@ -55,4 +55,4 @@ jobs:
       run: |
         cd wrappers && RUSTFLAGS="-D warnings" cargo clippy --all --tests --no-deps --features all_fdws,helloworld_fdw
 
-    - run: cd wrappers && cargo pgrx test --features all_fdws,pg15
+    - run: cd wrappers && cargo pgrx test --features "all_fdws pg15"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -387,6 +387,29 @@ dependencies = [
 ]
 
 [[package]]
+name = "aws-sdk-cognitoidentityprovider"
+version = "1.11.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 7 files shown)


burmecia:
test failed due to this ClickHouse issue: https://github.com/ClickHouse/ClickHouse/issues/60020

imor (approved):

PR #242, opened 2024-02-22
Add s3 endpoint support

encima:

Adds the option to provide `endpoint_url` when configuring an s3 server

Current behaviour does not allow this option and limits s3 functionality to AWS

## What is the new behavior?

Allows any s3 compatible providers to be used with the wrapper

## Additional context

Env var could not be used as it is [not supported in the SDK](https://github.com/awslabs/aws-sdk-rust/issues/932)

### DIFF
--- a/docs/s3.md
+++ b/docs/s3.md
@@ -100,6 +100,25 @@ We need to provide Postgres with the credentials to connect to S3, and any addit
       );
     ```
 
+The full list of options are below:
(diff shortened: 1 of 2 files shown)


imor (approved):
Worth updating docs as well?

encima:
> Worth updating docs as well?

Excellent point, thanks! Done with all options (and example)

imor (approved):

PR #255, opened 2024-04-16
Minor changes to options file

AayushyaVajpayee:
2 changes which are minor  refactor of the code.
1.  Use ok_or_else instead of ok_or. This change prevents the string allocation happening on every method call to only happen when there is an error.
2. SImplify nested if else block `check_option_contain` method. Also use format! macro to prevent to have only one string allocation instead of 2.

No change in behaviour
Please link any relevant issues here.

## What is the new behavior?
No change in behaviour
Feel free to include screenshots if it includes visual changes.

## Additional context

Add any other context or screenshots.

### DIFF
--- a/supabase-wrappers/src/options.rs
+++ b/supabase-wrappers/src/options.rs
@@ -57,7 +57,7 @@ pub fn require_option<'map>(
     options
         .get(opt_name)
         .map(|t| t.as_ref())
-        .ok_or_else(||OptionsError::OptionNameNotFound(opt_name.to_string()))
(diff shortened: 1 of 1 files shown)


imor:
@AayushyaVajpayee thanks for the PR. Could you please run `cargo fmt` to fix the formatting failure.

AayushyaVajpayee:
> @AayushyaVajpayee thanks for the PR. Could you please run `cargo fmt` to fix the formatting failure.

thanks. fixed formattting

imor (approved):

PR #263, opened 2024-05-14
fix: code improvements as per clippy

burmecia:

This code is to improve code quality as per clippy suggestion.

N/A

## What is the new behavior?

N/A

## Additional context

N/A

### DIFF
--- a/wrappers/src/fdw/auth0_fdw/auth0_client/row.rs
+++ b/wrappers/src/fdw/auth0_fdw/auth0_client/row.rs
@@ -1,16 +1,8 @@
 use pgrx::JsonB;
 use serde::Deserialize;
-use serde::Serialize;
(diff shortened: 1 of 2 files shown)


wrappers/src/fdw/auth0_fdw/auth0_client/row.rs
```diff
@@ -5,6 +5,7 @@ use supabase_wrappers::prelude::Cell;
 use supabase_wrappers::prelude::Column;
 use supabase_wrappers::prelude::Row;
 
+#[allow(dead_code)]
```
imor, line 8: Why not remove it if it not used?

burmecia: okay, removed it.

imor (approved):

PR #265, opened 2024-05-15
Feat: Extend the fields for users in the Auth0 FDW

romaingrx:

This PR extends the Auth0User struct to include all fields that the Auth0 endpoint may send (see [Auth0 API Documentation](https://auth0.com/docs/api/management/v2/users/get-users#response-messages)).

N/A

## What is the new behavior?

The Auth0User struct now includes all possible fields, allowing us to use these fields in a foreign table and apply quals.

## Additional context

N/A


burmecia:
Thanks for the PR! That's excellent but ere are some Clippy complaints, could you resolve it?

burmecia (approved):
LGTM, thanks!

PR #281, opened 2024-05-31
feat: add timestamp with time zone cell type support

burmecia:

This PR is to add a new column type `timestamp with time zone` support. With this new column type, below FDWs can support the new data type.

- Airtable
- BigQuery
- MS SQL Server
- S3 parquet
- Logflare

Currently only `timestamp without time zone` is supported.

## What is the new behavior?

Support `timestamp with time zone` in the framework.

## Additional context

N/A

### DIFF
--- a/supabase-wrappers/src/interface.rs
+++ b/supabase-wrappers/src/interface.rs
@@ -83,27 +83,43 @@ impl fmt::Display for Cell {
             Cell::Date(v) => unsafe {
                 let dt =
                     fcinfo::direct_function_call_as_datum(pg_sys::date_out, &[(*v).into_datum()])
-                        .unwrap();
+                        .expect("cell should be a valid date");
                 let dt_cstr = CStr::from_ptr(dt.cast_mut_ptr());
-                write!(f, "'{}'", dt_cstr.to_str().unwrap())
+                write!(
+                    f,
+                    "'{}'",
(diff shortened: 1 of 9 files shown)


supabase-wrappers/src/interface.rs
```diff
@@ -94,6 +96,15 @@ impl fmt::Display for Cell {
                 let ts_cstr = CStr::from_ptr(ts.cast_mut_ptr());
                 write!(f, "'{}'", ts_cstr.to_str().unwrap())
             },
+            Cell::Timestamptz(v) => unsafe {
+                let ts = fcinfo::direct_function_call_as_datum(
+                    pg_sys::timestamptz_out,
+                    &[(*v).into_datum()],
+                )
+                .unwrap();
```
imor, line 104: Can we replace all calls to `unwrap` with `expect(<error message>)`, it becomes much easier to troubleshoot with a message. We recently replaced all `unwrap` calls in `pg_graphql` with `expect` after we failed to find the cause of a failure.

imor (approved):

supabase-wrappers/src/interface.rs
burmecia: sure, I've done this in `interface.rs` file for this PR. Will apply this to other files in another PR.

PR #289, opened 2024-06-20
fix: remove `unwrap()` from stats related functions

burmecia:

This PR is to remove `unwrap()` call from stats related functions. This also caused failure when using API to access foreign tables in some cases.

Currently, it is hard to troubleshot when the panic is caused by `unwrap()` call.

## What is the new behavior?

Replace `unwrap()` with `expect()` or remove it if possible.

## Additional context

N/A

### DIFF
--- a/wrappers/src/stats.rs
+++ b/wrappers/src/stats.rs
@@ -1,5 +1,6 @@
 use pgrx::{prelude::*, JsonB};
 use std::fmt;
+use supabase_wrappers::prelude::report_warning;
 
 // fdw stats table name
 const FDW_STATS_TABLE: &str = "wrappers_fdw_stats";
@@ -102,11 +103,13 @@ pub(crate) fn set_metadata(fdw_name: &str, metadata: Option<JsonB>) {
             updated_at = timezone('utc'::text, now())",
         get_stats_table()
     );
-    let _ = Spi::run_with_args(
+    if let Err(err) = Spi::run_with_args(
         &sql,
         Some(vec![
             (PgBuiltInOids::TEXTOID.oid(), fdw_name.into_datum()),
             (PgBuiltInOids::JSONBOID.oid(), metadata.into_datum()),
         ]),
-    );
+    ) {
+        report_warning(&format!("set fdw stats metadata failed: {}", err));
+    };
 }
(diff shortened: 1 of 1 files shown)


burmecia:
#290 is to fix the test failure

wrappers/src/stats.rs
```diff
@@ -102,12 +102,11 @@ pub(crate) fn set_metadata(fdw_name: &str, metadata: Option<JsonB>) {
...
-    Spi::run_with_args(
+    let _ = Spi::run_with_args(
         &sql,
         Some(vec![
             (PgBuiltInOids::TEXTOID.oid(), fdw_name.into_datum()),
             (PgBuiltInOids::JSONBOID.oid(), metadata.into_datum()),
         ]),
-    )
-    .unwrap();
+    );
```
imor, line 111: Maybe we should emit a warning in the logs here instead of ignoring the error?

imor (approved):

wrappers/src/stats.rs
burmecia: sure, fixed.

imor (approved):

dohooo:
@burmecia, thanks for your work. Where can I follow the updates of the new release?

burmecia:
We're preparing a new release and it should be ready soon, you can check it on [the release page](https://github.com/supabase/wrappers/releases). Just a note once the new version of Wrappers is release it sill needed to be included in [Supabase Postgres image](https://github.com/supabase/postgres) to available on Supabase platform, this will take some time.

dohooo:
@burmecia I can confirm this problem still happens. I've tried to upgrade our service to the latest version(15.1.1.76). But I still need to execute the SQL in the studio client; otherwise, the call via JS SDK will fail.

dohooo:
@burmecia Does this version(15.1.1.76) include this change?

burmecia:
> @burmecia Does this version(15.1.1.76) include this change? image

not yet, that `15.1.1.74-wrapper` tag was for integration testing, it cannot be used. But the latest [Wrappers v0.4.1 has just been merged](https://github.com/supabase/postgres/pull/1029), it will be available on the next release of Supabase Postgres.

PR #294, opened 2024-07-01
feat(stripe): add `api_version` server option

### LINKED ISSUE #291: The products table is empty if the stripe account API version is old
# Bug report

## Describe the bug

My stripe account is very old, and had the "default API version 2016-03-07" set in the developer console. I have three products defined via the Stripe dashboard but none of them showed up in the wrapped tables in Supabase on `select * from stripe.products`. If I created a product by inserting into the table, I could see it on select but none other. The newly created product also shows up on the Stripe dashboard list of products. In Postgres running the query in the `psql` CLI, no warnings or errors are shown, just an empty list.

To debug this, I did a bunch of Stripe API calls via `curl` on my command line, and one of the tests returned this error to me:

```
curl -G https://api.stripe.com/v1/products/search -u "sk_test_XXXX:" --data-urlencode query="active:'true'"
{
  "error": {
    "message": "Search is not supported on api version 2016-03-07. Update your API version, or set the API Version of this request to 2020-08-27 or greater.",
    "request_log_url": "https://dashboard.stripe.com/logs/req_4bPEPEO1nvuTtj?t=1718974843",
    "type": "invalid_request_error"
  }
}
```

Upgrading my Stripe account's API version (bottom of the screen at https://dashboard.stripe.com/developers) to the latest, 2024-04-10, resolved this error. Now queries from the wrapper in supabase return all my products.

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

1. Go to Stripe and set the API version to an old one (not even sure this is possible)
2. Run the query in supabase `select * from stripe.products`
3. empty list is returned

## Expected behavior

The query should always return a list, or show the stripe API error.

The wrapper should explicitly request the latest API by setting the `Stripe-Version` header as per https://docs.stripe.com/api/versioning since it depends on the search endpoint which is not available until the 2020-08-27 version.

## System information

- OS: [e.g. macOS, Windows]
- Browser (if applies) [e.g. chrome, safari]
- Version of supabase-js: [e.g. 6.0.2]
- Version of Node.js: [e.g. 10.10.0]

burmecia:

This PR is to add a new `api_version` foreign server option for Stripe FDW. Fix #291 .

The current Stripe API requests are always using latest Stripe API version, which can cause issues when requesting legacy Stripe API.

## What is the new behavior?

Add a new `api_version` foreign server option, so users can specify a specific Stripe API version number when creating a foreign server.

More details about Stripe API versioning: https://docs.stripe.com/api/versioning

## Additional context

N/A

### DIFF
--- a/docs/stripe.md
+++ b/docs/stripe.md
@@ -45,7 +45,8 @@ We need to provide Postgres with the credentials to connect to Stripe, and any a
       foreign data wrapper stripe_wrapper
       options (
         api_key_id '<key_ID>', -- The Key ID from above, required.
-        api_url 'https://api.stripe.com/v1/'  -- Stripe API base URL, optional. Default is 'https://api.stripe.com/v1/'
+        api_url 'https://api.stripe.com/v1/',  -- Stripe API base URL, optional. Default is 'https://api.stripe.com/v1/'
+        api_version '2024-06-20'  -- Stripe API version, optional. Default is your Stripe account’s default API version.
       );
     ```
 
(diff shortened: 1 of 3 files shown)


imor (approved):

imor:
@vickkhera updated the description. More importantly, now users should be able to specify the Stripe API version they want to use via the new option. Does that work for you?

vickkhera:
@imor This is a workable solution. My solution was to just bump the version of the API in the Stripe dashboard, since I had nothing else using the API that might break with a newer version.

The docs https://supabase.github.io/wrappers/stripe/ should mention something about old versions being incompatible with this extension.

I still have this nagging feeling that the right thing in the clients using the API is to specify the version to which they are built, but I see no guidance from Stripe regarding that.

Thanks for following up.

PR #296, opened 2024-07-01
chore(notion): exclude notion fdw from next release

burmecia:

This PR is to exclude Notion FDW from next release, as it is not mature enough and we will rewrite it using the new Wasm framework.

N/A

## What is the new behavior?

Exclude Notion FDW just for now.

## Additional context

N/A


imor:
Was notion fdw released to customers before? If yes, what happens to projects which have notion fdw enabled?

PR #315, opened 2024-07-30
Change import_foreign_schema() signature.

remilapeyre:
`import_foreign_schema()` is now a method instead of static method that receives the server id.

### DIFF
--- a/supabase-wrappers/src/instance.rs
+++ b/supabase-wrappers/src/instance.rs
@@ -2,17 +2,6 @@ use crate::prelude::*;
 use pgrx::pg_sys::panic::ErrorReport;
 use pgrx::prelude::*;
 
-// create a fdw instance from a foreign table id
-pub(super) unsafe fn create_fdw_instance_from_table_id<
-    E: Into<ErrorReport>,
-    W: ForeignDataWrapper<E>,
->(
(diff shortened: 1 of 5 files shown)


supabase-wrappers/src/instance.rs
```diff
@@ -2,12 +2,25 @@ use crate::prelude::*;
...
+>(
     ftable_id: pg_sys::Oid,
 ) -> W {
     let ftable = pg_sys::GetForeignTable(ftable_id);
-    let fserver = pg_sys::GetForeignServer((*ftable).serverid);
+    create_fdw_instance_from_server_id((*ftable).serverid)
+}
+
+// create a fdw instance from its id
+pub(super) unsafe fn create_fdw_instance_from_server_id<
```
burmecia, line 17: can we put `create_fdw_instance_from_server_id` definition before `create_fdw_instance_from_table_id`? Just to make it more natural for reading.

burmecia:
That looks great! Thanks for the PR.

supabase-wrappers/src/instance.rs
remilapeyre: 👍

burmecia (approved):

PR #323, opened 2024-08-05
Give the full server struct to the FDW when instanciating it

remilapeyre:
The complete SQL command to create a foreign data wrapper is 

```
CREATE SERVER [ IF NOT EXISTS ] server_name [ TYPE 'server_type' ] [ VERSION 'server_version' ]
    FOREIGN DATA WRAPPER fdw_name
    [ OPTIONS ( option 'value' [, ... ] ) ]
```

At this time only the options are given to the Foreign Data Wrapper implementation. This patch changes the signature of `ForeignDataWrapper::new()` to give it the full structure that matches PostgreSQL ForeignServer to expose the server_name, server_type and server_version attributes.

### DIFF
--- a/supabase-wrappers/src/instance.rs
+++ b/supabase-wrappers/src/instance.rs
@@ -1,6 +1,16 @@
+use std::collections::HashMap;
+use std::ffi::CStr;
+
 use crate::prelude::*;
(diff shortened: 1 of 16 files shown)


burmecia (approved):
Great work, thanks for the PR!

Issue #318, opened 2024-07-31
Compilation error - mismatched types


glib-0:
# Bug report

I'm trying to compile the mssql_fdw wrapper on a Raspberry Pi 4 and encountering an issue when the compiler gets to supabase-wrappers. 

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug
```
error[E0308]: mismatched types
    --> supabase-wrappers/src/import_foreign_schema.rs:84:18
      |
      |             ret.push(command.as_pg_cstr());
      |                 ---- ^^^^^^^^^^^^^^^^^^^^ expected `*mut i8`, found `*mut u8`
      |                 |
      |                 arguments to this method are incorrect
      |
      = note: expected raw pointer `*mut i8`
                 found raw pointer `*mut u8`

```

## To Reproduce

1. Install Postgres 15
2. Install pgrx-0.11.3
3. Run `cargo pgrx install --no-default-features --features pg15,mssql_fdw --release` in ~/wrappers/wrappers

## Expected behavior
Build completes
## Screenshots
N/A
## System information
- Raspberry Pi 4B
- OS: Raspberry Pi OS March 2024
- Browser (if applies) N/A
- Version of supabase-js: N/A
- Version of Node.js: N/A

remilapeyre:
The return type of `as_pg_cstr()` changes based on the platform. I thought I had fixed that. The previous line should be `    let mut ret: PgList<ffi::c_char> = PgList::new();`

PR #324, opened 2024-08-06
Fix type of return value for import_foreign_schema()

remilapeyre:
I checked that this fixes the issue on ARM Linux by using a Docker container on a M2 Mac.

Closes https://github.com/supabase/wrappers/issues/318.


burmecia (approved):
thanks for the PR!

PR #359, opened 2024-10-31
feat: [DATAENG-986] add DuckDB foreign data wrapper

burmecia:

This PR is to add DuckDB foreign data wrapper.

There is no DuckDB foreign data wrapper.

## What is the new behavior?

The DuckDB FDW is added with below major features:

- Supported data source:
  - S3 and S3-compatible storage, such like Cloudflare R2 and Supabase Storage
  - Apache Iceberg services
    - S3 Tables
    - Cloudflare R2 Data Catalog
    - Apache Polaris
    - Lakekeeper
- Support `import foreign schema`
- Support `where`, `order by` and `limit` pushdown

## Additional context

N/A


docs/catalog/duckdb.md
```diff
@@ -0,0 +1,783 @@
...
+  );
+```
+
+Then, import all the tables in `docs_example` namespace and query it:
+
+```sql
+import foreign schema "docs_example"
+  from server duckdb_server into duckdb;
+
+select * from duckdb.s3_tables_docs_example_guides;
```
imor, line 742: Can we mention in the docs how these names are derived. E.g. `<type>_<import_foreign_schema_name>_<table_name>` etc.

PR #377, opened 2024-12-04
chore: generic error message for file load

staaldraad:

Bug fix

When loading a wasm file (from the local filesystem), the error message could be used to enumerate files on the filesystem.

File exists - `/etc/passwd`
```
wrappers=# select * from wasm_out;
ERROR:  wasmtime error: failed to parse WebAssembly module
```

No such file - `/etc/nop`
```
wrappers=# select * from wasm_out;
ERROR:  wasmtime error: failed to read input file: /etc/nop
```

## What is the new behavior?

Gives a generic error message for file load failures to prevent file system enumeration.

File exists - `/etc/passwd`
```
wrappers=# select * from wasm_out;
ERROR:  wasmtime error: failed to parse WebAssembly module
```

No such file - `/etc/nop`
```
wrappers=# select * from wasm_out;
ERROR:  wasmtime error: failed to parse WebAssembly module
```
## Additional context

Add any other context or screenshots.

### DIFF
--- a/wrappers/src/fdw/wasm_fdw/mod.rs
+++ b/wrappers/src/fdw/wasm_fdw/mod.rs
@@ -14,6 +14,9 @@ use self::bindings::supabase::wrappers::types::FdwError as GuestFdwError;
 
 #[derive(Error, Debug)]
 enum WasmFdwError {
+    #[error("invalid WebAssembly component")]
+    InvalidWasmComponent,
+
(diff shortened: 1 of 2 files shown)


burmecia (approved):
Thanks for the PR! I've made the error specific for the Wasm component load failure, so its error message won't leak filesystem info.

Issue #378, opened 2024-12-05
WASM FDW url resolution issue


akacase:
# Bug report
I have a few custom FDW's written with the postgres_fdw template provided. I've been working on an interesting error:

`Query 1 ERROR at Line 1: : ERROR:  wasmtime error: WebAssembly translation error`

There seems to be some incorrect logic in handling http(s) related remote connections.

https://github.com/supabase/wrappers/blob/8a30794da96da18b013beccfa7a957a998689a1d/wrappers/src/fdw/wasm_fdw/wasm_fdw.rs#L46-L122

## Describe the bug

Remote supabase postgres (15.6) database issues a: `Query 1 ERROR at Line 1: : ERROR:  wasmtime error: WebAssembly translation error` after defining a remote server and foreign table pointing at a https file hosted outside of the "release process" on GitHub actions. Local postgres (16.6) crashes.

## To Reproduce

first test:
1) compile fdw: `cargo component build --release --target wasm32-unknown-unknown`
2) create foreign wrapper with said output (I know checksum calculations aren't used for `file:///`)
3) test and works

second test:
1) compile fdw: ` cargo component build --release --target wasm32-unknown-unknown`
2) serve the file via a simple Python server locally or in my case Deno (2.1.2).
3) test and works `http://localhost:8000/fdw/test_fdw.wasm`

third test:
1) stick compiled fdw anywhere; on a supabase S3 bucket,  apache httpd server, even a public release on GitHub
after compiling locally via `gh release`
2) create server, foreign table and see error prop up.

fourth test:
1) use local postgres
2) point to remotely hosted wasm
3) create server and table
4) postgres crashes

behavior is also present (no database crash) but WASM error on Postgres 15.6 (hosted on Supabase)

This reliably works on all supabase wasm-fdw's built locally with: `cargo component build --release --target wasm32-unknown-unknown` and served via any of the aforementioned endpoints: paddle, notion, etc.

```sh
case@corp paddle_fdw % wasm-tools validate -vv target/wasm32-unknown-unknown/release/paddle_fdw.wasm 
[2024-12-05T18:31:28Z INFO ] read module in 733.167µs
[2024-12-05T18:31:28Z INFO ] module structure validated in 3.883958ms
[2024-12-05T18:31:28Z INFO ] functions validated in 1.401333ms
```

```sh
case@corp paddle_fdw % curl -I https://wiw.org/~case/paddle_fdw.wasm   
HTTP/1.1 200 OK
[...]

PR #379, opened 2024-12-06
fix(wasm): clean up download, byte handling and hashing

akacase:

#378 

#378 

## What is the new behavior?

The wasmtime errors were _probably_ because:
- The file was incomplete (partial download)
- The file was corrupted during the save
- The file wasn't properly closed/flushed after writing

These issues might not have appeared until recently because:
- Those servers might have better-behaved HTTP responses
- The download paths might have been more thoroughly tested

The new code is more defensive about:
- Complete downloads
- Proper byte handling
- Atomic file operations
- Error propagation

## Additional context

I have added bytes here for a specific reason:

The `bytes` crate is specifically designed for working with raw binary data
It ensures we don't accidentally corrupt binary data through string conversions or character encoding issues. This is crucial for WASM files, which must maintain exact byte-level integrity. Bytes itself is highly efficient (zero-copy).

Thanks!


akacase:
fixed logs leaking, we _should_ look into a log framework, where debug level logs can be used and then those paths are not valid in production.

akacase:
another failure mode is cache permission, this would be more ideal:

```rust
   let mut path = std::env::var("PGDATA")
        .map(PathBuf::from)
        .map_err(|_| "PGDATA environment variable not set")?
        .join("extension_cache")
        .join("wasm_fdw");
```

Three issues I see with the current approach:
- `dirs::cache_dir()` uses the system's default cache directory
- Could lead to permission problems when Postgres is running as a different user
- Cache poisoning, because files are being stored in a user-writable location rather than a Postgres-controlled directory

burmecia:
Thanks for the PR! All the improvements look good to me but I am not much in favour of using `PGDATA` as the cache dir. The `PGDATA` dir is mainly used by Postgres internally and stored the core files and folders. IMO Wrappers as an extension should not touch the internal Postgres storage, rather it is better to use cache dir outside Postgres core system.

`dirs::cache_dir()` is not the system's default cache directory, it is specific to current user which is the user running Postgres process. For its [documentation](https://docs.rs/dirs/latest/dirs/fn.cache_dir.html),

Platform | Value | Example
-- | -- | --
Linux | $XDG_CACHE_HOME or $HOME/.cache | /home/alice/.cache
macOS | $HOME/Library/Caches | /Users/Alice/Library/Caches
Windows | {FOLDERID_LocalAppData} | C:\Users\Alice\AppData\Local

So the cache dir is still owned by Postgres, and normally it is not writable by other users.

akacase:
> Thanks for the PR! All the improvements look good to me but I am not much in favour of using `PGDATA` as the cache dir. The `PGDATA` dir is mainly used by Postgres internally and stored the core files and folders. IMO Wrappers as an extension should not touch the internal Postgres storage, rather it is better to use cache dir outside Postgres core system.
> 
> `dirs::cache_dir()` is not the system's default cache directory, it is specific to current user which is the user running Postgres process. For its [documentation](https://docs.rs/dirs/latest/dirs/fn.cache_dir.html),
> 
> Platform	Value	Example
> Linux	$XDG_CACHE_HOME or $HOME/.cache	/home/alice/.cache
> macOS	$HOME/Library/Caches	/Users/Alice/Library/Caches
> Windows	{FOLDERID_LocalAppData}	C:\Users\Alice\AppData\Local
> So the cache dir is still owned by Postgres, and normally it is not writable by other users.

understood, thanks!

akacase:
I'll hold back on the patch for cache path, but all the commits here I believe are good to go. Let me know if you need anything else.

burmecia:
@akacase I've made some minor changes on top of your patch, please have a look at your repo and review it. If everything is fine, we will be ready to merge. Thanks.

akacase:
merged, thanks!

burmecia (approved):

PR #393, opened 2024-12-18
updated airtable docs

TheOtherBrian1:

docs update

Some developers have expressed confusion over how to configure the airtable wrapper.
- [Example 1](https://app.hubspot.com/live-messages/19953346/inbox/8182677130#email)
- [Example 2](https://app.hubspot.com/live-messages/19953346/inbox/8519251499)

They're mainly confused about how to obtain the base_id and table_id. Another issue is that the wrapper only works when the name is lowercase. 

## What is the new behavior?

Added above caveats


burmecia (approved):
Thanks for the PR!

PR #404, opened 2025-01-24
feat(wasm): add Clerk Wasm FDW

burmecia:

This PR is to add a new [Clerk](https://clerk.com/) Wasm foreign data wrapper.

N/A

## What is the new behavior?

N/A

## Additional context

N/A


burmecia:
CI failed due to https://github.com/pgcentralfoundation/pgrx/issues/1966

burmecia:
Need to wait for #409 to be merged first to fix CI issue.

colinclerk:
Hello - thank you for putting this together! I read through the PR and we're going to suggest a few adjustments (change table name to `clerk_{object}`, and cherry pick a more complete column set, e.g. I know we'd like to get primary_email_address in there directly instead of having to loop through attrs).

Would you prefer we fork this branch or another process? Since we'd plan on changing the table names, everything we suggest will be backward compatible, so there's no need to block on us if you want to move forward.

Thanks again!

burmecia:
Thanks for your suggestions @colinclerk.

We already suggested using dedicated schema `clerk` to save foreign tables, so I don't think table name prefix like `clerk_{object}` is necessary, but that's all up to the user, they can use whatever names they like and we just provide recommendations to better organise the tables.

For cheery picking column set, we definitely can pick some more properties if you think they are more important, but that only works for 1-to-1 properties. For some properties like email addresses, they are 1-to-many relationship with the object, and currently the foreign table can only return 1 row for 1 object, so we'd better to keep those properties in json column and let user use subquery or view to extract it out.

You can directly push on this branch or create another fork, we'd like to see more feedbacks from you for this fdw and thanks again!

imor (approved):

wobsoriano:
Hi @burmecia, I saw the recent release that includes this update and immediately tried it out. It works perfectly! Thank you for implementing this 🫡

PR #424, opened 2025-02-12
fix(mssql): numeric data type conversion precision loss

### LINKED ISSUE #416: Numeric data type in MS sql querry wrapper
# Bug report

- [x ] I confirm this is a bug with Supabase, not with my own application. 
- [x ] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

I have an issue regarding FDW wrapper for MS SQL server.
Numeric type dosen't work correctly it bypasses any decimals only provides rounded numer f.e (instead of 3.20 it delivers 3)

SQL query for table generation:
```
create foreign table Bible
(
    id           int    not null,
    kaina     numeric(18,2)   not null
)
  server fuck_me_server
  options ( table 'dbo.detales' );
```

On the MS SQL server side i have already Decimal (18, 2 ) value
Image

Query output on SQL server:

Image

Querry output on postgresql

Image

## System information

Official docker image of supabase-postgress installed via CLI
OS Windows 11

burmecia:

This PR is to fix the numeric data type conversion precision loss issue in MSSQL FDW, fix #416.

The `numeric` data type is converted to integer, which caused precision loss.

## What is the new behavior?

Use `f64` as the converted data type, so to keep number precision.

## Additional context

N/A

### DIFF
--- a/wrappers/src/fdw/mssql_fdw/README.md
+++ b/wrappers/src/fdw/mssql_fdw/README.md
@@ -10,6 +10,7 @@ This is a foreign data wrapper for [Microsoft SQL Server](https://www.microsoft.
 
 | Version | Date       | Notes                                                |
 | ------- | ---------- | ---------------------------------------------------- |
+| 0.1.3   | 2025-02-12 | Fix Numeric type conversion error                    |
 | 0.1.2   | 2024-09-30 | Support for pgrx 0.12.6                              |
 | 0.1.1   | 2024-09-09 | Add boolean test qual support                        |
 | 0.1.0   | 2023-12-27 | Initial version                                      |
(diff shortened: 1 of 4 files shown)


imor:
All `numeric` values can't fit in an `f64`. An arbitrary precision crate like [rust-decimal](https://github.com/paupino/rust-decimal) is more appropriate imo.

burmecia:
The reason using `f64` here is because the cell encapsulates `pgrx::AnyNumeric` to represent a `numeric` type, but that `pgrx::AnyNumeric` can only be converted from a Rust primitive types such like i32, i64, f64 and etc.

imor (approved):

PR #443, opened 2025-04-10
docs: [DATAENG-916] update Vault encryption method

burmecia:

It is now recommended to use `vault.create_secret` to save credential in Vault, this PR is to update all wrappers docs to use that function rather than using `insert into vault.secrets`.

N/A

## What is the new behavior?

N/A

## Additional context

N/A


burmecia:
coverage check failure due to disk space not enough, will fix it in another PR.

imor (approved):

PR #462, opened 2025-05-18
feat: [PSQL-367] add Apache Iceberg FDW

burmecia:

This PR is to add Apache Iceberg FDW.

There is no Apache Iceberg FDW.

## What is the new behavior?

Apache Iceberg FDW is added, for now it supports:

- AWS S3 tables
- REST Catalog + S3 storage

## Additional context

- There are some defects as below needs to be fixed in upstream `iceberg-rust`, so for now we just use a patched version of [`iceberg-rust`](https://github.com/burmecia/iceberg-rust)
  - https://github.com/apache/iceberg-rust/pull/1347
  - https://github.com/apache/iceberg-rust/pull/1346
- `iceberg-catalog-s3tables` crate is not officially released yet in upstream, so we use it directly from upstream source code
- Because of the new dependency added, some code refactoring is also added:
  - Rust version upgraded to `1.85.1`
  - S3 FDW removed code for mocking
  - [`rustls`](https://docs.rs/rustls/latest/rustls/index.html#cryptography-providers) upgraded to `0.23.27` which requires to specify cryptographic primitives provider at start, so we added a function `setup_rustls_default_crypto_provider` to do that and for now only BigQuery FDW needs it
  - created a sub-workspace for Wasm wrappers and unified Rust version and edition spec
  - replaced `local-stack` with `minio` container, and shared it in S3 and Iceberg FDW testing
  - other minor CI adjustments to adopt all the above changes


Cargo.lock
```diff
@@ -3310,23 +3583,109 @@ dependencies = [
  "cc",
 ]
 
+[[package]]
+name = "iceberg"
+version = "0.4.0"
+source = "git+https://github.com/burmecia/iceberg-rust?branch=main#a1eaf8669c818da89ce9d7fa2bccdf162767ae74"
```
imor, line 3589: Are we going to upstream any changes made in this fork?

burmecia: yes, already raised PR to upstream:

- https://github.com/apache/iceberg-rust/pull/1347
- https://github.com/apache/iceberg-rust/pull/1346

docs/catalog/iceberg.md
```diff
@@ -0,0 +1,323 @@
...
+## Query Pushdown Support
+
+This FDW supports `where` clause pushdown with below operators.
+
+| Operator                                   | Note                             |
+| ------------------------------------------ | -------------------------------- |
+| `=`, `>`, `>=`, `<`, `<=`, `<>`, `!=`      |                                  |
+| `is null`, `is not null`                   |                                  |
+| `x`, `not x`, `x is true`, `x is not true` | column `x` data type is `boolean`|
+| `x between a and b`                        | column `x` data type is `date` or `timestamp` |
```
imor, line 235: Isn't this supported for numeric types like float, int, double etc.?

wrappers/src/fdw/iceberg_fdw/pushdown.rs
```diff
@@ -0,0 +1,196 @@
...
+        Cell::Bool(v) => Some(Datum::bool(*v)),
+        Cell::F32(v) => match tgt_type {
+            Type::Primitive(PrimitiveType::Float) => Some(Datum::float(*v)),
+            Type::Primitive(PrimitiveType::Double) => Some(Datum::double(*v)),
+            _ => None,
+        },
+        Cell::I32(v) => match tgt_type {
+            Type::Primitive(PrimitiveType::Int) => Some(Datum::int(*v)),
+            Type::Primitive(PrimitiveType::Long) => Some(Datum::long(*v)),
+            Type::Primitive(PrimitiveType::Float) => Some(Datum::float(*v as f32)),
```
imor, line 35: There's some precision loss here when converting from an `i32` to an `f32` as not all `i32` values can be represented by an `f32`.

wrappers/src/fdw/iceberg_fdw/pushdown.rs
```diff
@@ -0,0 +1,196 @@
...
+            Type::Primitive(PrimitiveType::Double) => Some(Datum::double(*v)),
+            _ => None,
+        },
+        Cell::F64(v) => match tgt_type {
+            Type::Primitive(PrimitiveType::Float) => Some(Datum::float(*v as f32)),
```
imor, line 40: Same here, precision loss.

docs/catalog/iceberg.md
burmecia: yes, numeric types are supported

docs/catalog/iceberg.md
```diff
@@ -0,0 +1,323 @@
...
+#### Operations
+
+| Object     | Select | Insert | Update | Delete | Truncate |
+| ---------- | :----: | :----: | :----: | :----: | :------: |
+| table      |   ✅   |   ❌   |   ❌   |   ❌   |    ❌    |
+
+#### Usage
+
+```sql
+create foreign table iceberg.guides (
```
olirice, line 214: it would be good to call our that this is only necessary if you did not use import foreign schema

PR #470, opened 2025-05-27
fix(clickhouse): integer type up-cast and down-cast

burmecia:

This PR is to fix some integer data type conversion issues between ClickHouse and Postgres.

Currently some of the integer data types conversion from ClickHouse -> Postgres:

- Bool -> missing
- Int8 -> missing
- UInt8 -> boolean
- Nullable(Bool) -> missing
- Nullable(Int8) -> missing
- Nullable(UInt8) -> boolean

## What is the new behavior?

Added some missing data type conversions:

From ClickHouse -> Postgres:

- Bool -> boolean
- Int8 -> "char"
- UInt8 -> smallint
- Nullable(Bool) -> boolean
- Nullable(Int8) -> "char"
- Nullable(UInt8) -> smallint

From Postgres -> ClickHouse, according to the target column type, we do below down casting:

- smallint -> UInt8
- integer -> UInt16
- bigint -> UInt32

## Additional context

N/A

### DIFF
--- a/docs/catalog/clickhouse.md
+++ b/docs/catalog/clickhouse.md
@@ -164,7 +164,9 @@ This FDW supports `where`, `order by` and `limit` clause pushdown, as well as pa
 
 | Postgres Type      | ClickHouse Type   |
 | ------------------ | ----------------- |
-| boolean            | UInt8             |
+| boolean            | Bool              |
+| "char"             | Int8              |
+| smallint           | UInt8             |
 | smallint           | Int16             |
 | integer            | UInt16            |
 | integer            | Int32             |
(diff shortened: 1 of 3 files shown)


imor:
> From Postgres -> ClickHouse, according to the target column type, we do below down casting:
> 
> smallint -> UInt8
> integer -> UInt16
> bigint -> UInt32

Postgres smallint is a 2 byte signed integer, shouldn't it be mapped to ClickHouse Int16 type which is also a 2 byte signed integer? Same for integer (4 bytes signed) to Int32 (4 bytes signed) and bigint (8 bytes signed) to In64 (8 bytes signed).

imor:
Ah cool, makes sense.

PR #477, opened 2025-06-15
fix(iceberg): refactor sever options passdown

burmecia:

This PR is mainly to refactor how the server options passed down to iceberg crate.

Currently, only a few hand-picked options in `create server` are passed down to iceberg crate. Those options are not comprehensive and have unnecessary name translation.

## What is the new behavior?

All the options in `create server` are passed down to iceberg crate without modification, except the option has `vault_` prefix, which will be replaced by decrypted text from Vault and the name striped that prefix. With this change, we can support any the iceberg services as upstream supported.

Other changes in this PR:

- As upstream iceberg crate just released version `0.5.1`, we switched to use it
- Upstream S3 tables catalog crate hasn't released yet, so we used a fork of it
- Updated docs and added one more example `Read Cloudflare R2 Data Catalog`

## Additional context

N/A

### DIFF
--- a/docs/catalog/iceberg.md
+++ b/docs/catalog/iceberg.md
@@ -345,9 +345,9 @@ create server iceberg_server
   options (
     aws_access_key_id '<R2_access_key_ID>',
     aws_secret_access_key '<R2_secret_access_key>',
-	token '<R2 API token>',
-	warehouse 'xxx_r2-data-catalog-tutorial',
-	"s3.endpoint" 'https://xxx.r2.cloudflarestorage.com',
+    token '<R2 API token>',
+    warehouse 'xxx_r2-data-catalog-tutorial',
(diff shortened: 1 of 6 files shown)


docs/catalog/iceberg.md
```diff
@@ -308,6 +335,31 @@ Then query the foreign table:
...
+
+```sql
+create server iceberg_server
+  foreign data wrapper iceberg_wrapper
+  options (
+    aws_access_key_id '<R2_access_key_ID>',
+    aws_secret_access_key '<R2_secret_access_key>',
+	token '<R2 API token>',
+	warehouse 'xxx_r2-data-catalog-tutorial',
+	"s3.endpoint" 'https://xxx.r2.cloudflarestorage.com',
```
kiwicopple, line 350: possible tabs issue here

kiwicopple (approved):

PR #493, opened 2025-08-14
fix(airtable): make column name case insensitive

### LINKED ISSUE #171: Uppercased columns for Airtable FDW never return a value
# Bug report

## Describe the bug

Quoted column names with uppercase letters do not return a value for Airtable wrapper.

## To Reproduce

1. Create a foreign data table with a quoted column name with an uppercase letter:

```sql
create foreign table my_foreign_table (
  "Full Name" text
)
server airtable_server
options (
  base_id 'appXXXX',
  table_id 'tblXXXX'
);
```

2. Run `select * from my_foreign_table`.
3. Note that the `Full Name` column never returns a value.

## Expected behavior

The `Full Name` column should return a value.

## Screenshots

N/A

## System information

Reproducible on [latest wrappers ref](https://github.com/supabase/wrappers/commit/f256e95843a934fe75633a16adbc357f97ec95dd).
 
## Additional context

N/A

burmecia:

This PR is to make column name case insensitive in Airtable wrapper which is same as Airtable. Fix #171.

The column name in foreign table must be lowercase, otherwise it won't be populated if the column name is uppercase in Airtable.

## What is the new behavior?

The column name in foreign table is now case insensitive, this is done by converting column name to lowercase from both sides and then match them together.

## Additional context

N/A

### DIFF
--- a/wrappers/dockerfiles/airtable/server.py
+++ b/wrappers/dockerfiles/airtable/server.py
@@ -41,7 +41,7 @@ if __name__ == "__main__":
     client.create(
         test_table,
         {
-            "bool_field": True,
+            "BOOL_field": True,
             "numeric_field": 1,
             "string_field": "two",
             "timestamp_field": "2023-07-19T06:39:15.000Z",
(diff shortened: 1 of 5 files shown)


imor (approved):

burmecia:
@gaoag we will make a new release soon, thanks.

PR #511, opened 2025-10-17
Adds MotherDuck support to DuckDB native fdw.

scottynomad:

Adds support for MotherDuck as a remote server type in the DuckDB fdw.

MotherDuck is not currently supported.

## What is the new behavior?

```
create server md_server
  foreign data wrapper duckdb_wrapper
  options (
    type 'md',
    database 'my_db',
    motherduck_token 'eyJ...cU'
  );
create schema if not exists md_local;
import foreign schema "main" limit to ("table1", "table2") from server md_server into md_local;
```

or using vault secrets
```
create server md_server
  foreign data wrapper duckdb_wrapper
  options (
    type 'md',
    database 'my_db',
    vault_motherduck_token 'secret_id'
  );
```

## Additional context

MotherDuck is remote DuckDB instance that gets ATTACHed to the local one so we're able to honor the LIMIT TO / EXCEPT table list arguments to `IMPORT FOREIGN SCHEMA` by pushing them down to the query agains the remote information_schema.

MotherDuck doesn't use a DuckDB secret block, instead it relies on env vars (not relevant here) or configuration settings. Despite what the docs say, passing those settings as url params to the first argument to ATTACH doesn't seem to work any longer.

The test coverage is limited at the moment because afaik there isn't an emulator for md. One potential improvement would be to also add support for attaching to a remote duckdb file on s3 which would share most of this functionality other than the configuration logic. i.e. https://duckdb.org/docs/stable/guides/network_cloud_storage/duckdb_over_https_or_s3

Happy to update the docs accordingly if this change is workable.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -40,7 +40,7 @@ jobs:
           sudo apt -y autoremove && sudo apt -y clean
           sudo chmod a+rwx `/usr/lib/postgresql/15/bin/pg_config --pkglibdir` `/usr/lib/postgresql/15/bin/pg_config --sharedir`/extension /var/run/postgresql/
 
-      - run: cargo install cargo-pgrx --version 0.16.0
+      - run: cargo install --locked cargo-pgrx --version 0.16.0
       - run: cargo pgrx init --pg15 /usr/lib/postgresql/15/bin/pg_config
 
       - name: Build docker images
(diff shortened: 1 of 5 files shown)


burmecia:
Sorry for the delayed CI run, the clippy seems still not happy, could you take a look again?

scottynomad:
> Sorry for the delayed CI run, the clippy seems still not happy, could you take a look again?

You're right @burmecia, sorry I missed that. I just pushed that fix. Also went on a journey to get vault installed locally and tested that the vault-backed config works.

PR #561, opened 2026-01-20
fix(s3): add delimiter option support for CSV files

7ttp:
## Summary

Added support for custom CSV delimiter option in S3 FDW to handle CSV files with non comma delimiters.

## Problem

S3 FDW only supported comma delimited CSV files. Users couldn't read CSV files with other delimiters like semicolons, tabs, or pipes.

## Solution
- Added csv_delimiter field to S3Fdw struct (defaults to comma)
- Parse delimiter option from foreign table options during begin_scan
- Pass custom delimiter to CSV reader builder when creating parser
 
## Related
- Closes https://github.com/supabase/supabase/issues/42004


burmecia:
Thanks for the PR 👍 ! I've added test case and docs for the new `delimiter` option.

burmecia (approved):

PR #573, opened 2026-02-17
feat(openapi): upgrade OpenAPI FDW to v0.2.0 with modular architecture

codybrom:
This is the v0.2.0 upgrade to the OpenAPI FDW, following the initial v0.1.x that landed in #566.

The biggest change is structural. The single ~1100-line `lib.rs` has been broken into focused modules (`config`, `request`, `response`, `pagination`, `column_matching`, `spec`, and `schema`), each with co-located tests. The original was doing too many things in one file (parsing specs, making HTTP requests, matching columns, handling pagination, etc.), which made it hard to follow and harder to maintain. Splitting it up makes each piece easier to test and understand on its own.

#### YAML spec support

OpenAPI specs published as YAML are now parsed automatically. `spec_url` accepts both JSON and YAML specs — the FDW tries JSON first, then falls back to YAML. This matters because some APIs only publish YAML specs.

#### Real-world examples

Five working examples showing how to use the FDW against real APIs: [NWS Weather](https://github.com/codybrom/wrappers/tree/openapi-fdw-v0.2.0/wasm-wrappers/fdw/openapi_fdw/examples/nws), [CarAPI](https://github.com/codybrom/wrappers/tree/openapi-fdw-v0.2.0/wasm-wrappers/fdw/openapi_fdw/examples/carapi), [PokeAPI](https://github.com/codybrom/wrappers/tree/openapi-fdw-v0.2.0/wasm-wrappers/fdw/openapi_fdw/examples/pokeapi), [GitHub](https://github.com/codybrom/wrappers/tree/openapi-fdw-v0.2.0/wasm-wrappers/fdw/openapi_fdw/examples/github), and [Threads](https://github.com/codybrom/wrappers/tree/openapi-fdw-v0.2.0/wasm-wrappers/fdw/openapi_fdw/examples/threads). Each has a README walkthrough and an `init.sql` you can run directly, including `IMPORT FOREIGN SCHEMA` quick-start sections.

#### POST endpoints now work for reads

v0.1 only supported GET endpoints. v0.2 adds POST-for-read, so APIs that use POST for search or filter operations (set `method 'POST'` and `request_body` on the table) can now work too.

#### Inline spec support

You can now embed an OpenAPI spec directly in the server options with `spec_json` instead of pointing to a URL. This is useful for APIs without a public OpenAPI spec (like Threads) or for customizing a subset of endpoints. The Threads example demonstrates this with a hand-written inline spec.

#### Smarter LIMIT handling and safety guards

On the query side, `LIMIT` is now forwarded to the API's page-size parameter, so `SELECT * FROM users LIMIT 10` won't fetch 100 rows just to throw away 90. There are also safety limits (`max_pages`, `max_response_bytes`) to prevent runaway pagination from burning through API quotas.

#### Automatic retries for transient errors

The FDW now retries on HTTP 429 (rate limit), 502 (bad gateway), and 503 (service unavailable) with exponential backoff, up to 3 retries. The 429 hint suggests reducing page size; 502/503 retries are silent.

#### Everything else

- API keys can be sent as query parameters or cookies (not just headers) via `api_key_location` server option
- Added a `debug 'true'` server option that logs request/response details as PostgreSQL INFO messages
- OpenAPI 3.1 support (type arrays, `[type, "null"]` nullable syntax, etc.)
- Header deduplication prevents duplicate `content-type` or `authorization` headers when custom headers overlap defaults
- Empty/whitespace-only credentials are filtered out with a warning instead of sending empty auth headers
- `RowsOut` stats now accurately count rows consumed by PostgreSQL (not just rows fetched)
- Catalog docs (`docs/catalog/openapi.md`) are updated for the new options but the structure follows the existing pattern

### Testing

531 unit tests across all modules, plus a Docker-based integration test setup (`test/run.sh`) with a mock API server and 113 assertions covering pagination, path parameters, error handling, and type coercion.

Five new real-world examples are included (Weather.gov, Threads, GitHub, CarAPI and PokeAPI). These double as documentation and a way to validate the FDW against actual APIs.

## Test plan

- [x] `cargo test` passes (531 unit tests, native target)
- [x] `cargo component build --release --target wasm32-unknown-unknown` produces valid wasm
- [x] `cargo fmt --check` and `cargo clippy` clean
- [x] Docker integration tests pass (`test/run.sh`)
- [x] Catalog docs render correctly

### DIFF
--- a/wasm-wrappers/fdw/openapi_fdw/src/config.rs
+++ b/wasm-wrappers/fdw/openapi_fdw/src/config.rs
@@ -135,6 +135,10 @@ impl ServerConfig {
 
         if let Some(user_agent) = user_agent {
             self.headers.push(("user-agent".to_owned(), user_agent));
+        } else {
+            // Default user-agent if not provided (some APIs reject requests without it)
+            self.headers
+                .push(("user-agent".to_owned(), "Wrappers OpenAPI FDW".to_string()));
         }
 
         if let Some(accept) = accept {


codybrom:
Hi @burmecia, this is ready for review. All CI checks are passing (including wasm integration tests) and I've addressed the Copilot and CodeRabbit feedback in the latest commit.

The main changes from v0.1.x (which you already approved in #566) are the modular refactor for maintainability and a few new capabilities: YAML/inline specs, POST-for-read, LIMIT pushdown, and OpenAPI 3.1 support. The core FDW interface and table options are backwards compatible.

Happy to walk through anything or make changes. Let me know!

PR #604, opened 2026-06-01
feat(wasm): per-request credentials from session variables

codybrom:

Feature. Adds a generic host function for reading a Postgres session variable at request time, plus an OpenAPI FDW option that uses it as the auth credential.

My use case is per-user access in a multi-tenant app. Each user has their own upstream token to a third-party system (OAuth access tokens, per-user API keys) and with Supabase Auth those tokens live in a per-user, RLS-protected table keyed to `auth.uid()`. My approach is a `SECURITY DEFINER` function that reads the calling user's token and pins it for the life of the query. This lines up with how Supabase already threads per-user context into Postgres. The `auth.uid()` comes from a session GUC (`request.jwt.claims`) so resolving a credential from a session variable is the same shape and not a new concept.

```sql
-- Tokens live in a per-user, RLS-protected table keyed to auth.uid()
-- SECURITY DEFINER reads the caller's token and pins it for the life of the query
create function set_api_token() returns void
  language sql security definer set search_path = '' as $$
  select set_config('app.api_token',
    (select access_token
       from public.user_integrations
      where user_id = auth.uid()),
    true);
$$;

-- per request:
select set_api_token();
select * from some_foreign_table;
```

A wasm FDW can only authenticate with a static credential that's fixed when the server is created, either inline in the server options or a Vault secret looked up by id/name. There's no way to hand it a credential that changes per request, per user, or per transaction.

## What is the new behavior?

With this PR, stored Supabase Auth credentials can be resolved from a Postgres session variable on each request, so it can match per user/request/transaction. Implemented with two main pieces that are purely additive and backward compatible so an absent or empty setting is a no-op and any static credential configured at the server level still applies.

1. `query-setting(name)` added to the utils WIT interface (v1 + v2), backed by `current_setting(name, true)` via SPI. Returns None when the setting is unset.
2. `auth_token_setting` / `auth_token_prefix` server options on the OpenAPI FDW. When `auth_token_setting` is set, the FDW reads that session variable per request and uses it as the Authorization token.

### Design note

As shipped, the query-setting function could read any Postgres session setting, not just an auth token, but the auth option is the only setting being read. I could have made it smaller and more focused as auth-only, but I went general because it's a simple tool other FDWs could reuse and other session settings aren't typically sensitive. Could be tightened to read auth only if preferred.

### Alternatives considered

| Approach | Notes |
|---|---|
| Static server option / Vault secret (today) | Vault resolves the secret once, not per query against session state, so it can't vary per user. |
| Token as a query qual (hidden column or `WHERE token = ...`), handled entirely in the guest | Doesn't need an interface change, but the token ends up in the SQL statement text and can leak into pg_stat_statements, logs, and EXPLAIN. Non-starter for a live credential. |
| Set per-user Vault secrets and have FDW pick the right one by name | Doesn't avoid the problem. FDW can't tell which user is calling (no auth.uid() inside the sandbox) without a per-query signal and it's a lot of secrets to create and rotate. |
| Session-GUC host function (what I built) | Optional additive change to utils WIT interface, no-ops when the setting isn't there (static creds untouched), keeps the secret out of the query text since it's a set_config session var and is generic enough (hopefully) that any other wasm FDW could use it. |

### Testing

- Unit tests for `apply_session_token` covering the edge cases (empty/whitespace no-op, custom prefix, replacing vs appending the header, overriding a static credential).
- A pg_test (`openapi_session_token_injection`) that runs the full path against the mock server: with the GUC unset no Authorization header is sent, and after set_config the token arrives prefixed.

### DIFF
--- a/docs/catalog/openapi.md
+++ b/docs/catalog/openapi.md
@@ -105,6 +105,8 @@ We need to provide Postgres with the credentials to access the API and any addit
 | `api_key_location` | No | Where to send the API key: `header` (default), `query`, or `cookie`. |
 | `bearer_token` | No | Bearer token for authentication (alternative to `api_key`). |
 | `bearer_token_id` | No | Vault secret key ID storing the bearer token. |
+| `auth_token_setting` | No | Name of a Postgres session variable (GUC) to read the auth token from at request time, e.g. `app.api_token`. When set and non-empty it overrides any static credential for that request. See [Per-request credentials](#per-request-credentials-session-variables). |
+| `auth_token_prefix` | No | Prefix for the `auth_token_setting` value in the Authorization header (default: `Bearer`). Set to an empty string to send the raw token. |
 | `user_agent` | No | Custom User-Agent header value. |
 | `accept` | No | Custom Accept header for content negotiation (e.g., `application/geo+json`). |
 | `headers` | No | Custom headers as JSON object (e.g., `'{"X-Custom": "value"}'`). |
@@ -402,7 +404,7 @@ options (endpoint '/users');
 
 - **Read-only**: This FDW only supports SELECT operations. INSERT, UPDATE, and DELETE are not supported at this time.
 - **No transactions**: Each SQL statement results in immediate HTTP requests; there is no transactional grouping.
-- **Authentication**: Currently supports API Key and Bearer Token authentication. OAuth flows are not supported.
+- **Authentication**: Supports API Key and Bearer Token authentication, either static (server option or Vault) or resolved per request from a session variable (see [Per-request credentials](#per-request-credentials-session-variables)). The FDW does not run OAuth flows itself, but a session variable lets you supply a token your application already obtained.
 - **OpenAPI version**: Only OpenAPI 3.0+ specifications are supported (not Swagger 2.0).
 
 ## Automatic Retries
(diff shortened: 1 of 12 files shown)


codybrom:
Addressed all four Copilot review comments in `5cd7c01`:

1. **Case-insensitive authorization match**: The headers server option isn't normalized, so a static `Authorization` header could've been duplicated instead of replaced. Switched the lookup to `eq_ignore_ascii_case` and added a regression test with a capitalized Authorization header.
2. **Misleading test name**: renamed to `test_session_token_prefix_default_empty` and reworded the comment so it matches what it asserts.
3. **Swallowed SPI error in query_setting**: now reports the error before returning `None`, matching `get_vault_secret` right above it. I kept the `Option` return rather than going to `Result`, since the WIT host function is `option<string>`. Undefined settings still come back as `None` via `current_setting(name, true)` so only unexpected SPI failures will get reported now.
4. **Empty/whitespace auth_token_setting**: trimmed to `None` at parse so the request path doesn't fire a `current_setting('')` lookup and trimmed `auth_token_prefix` too (to avoid a 'Bearer ' double-space case).

burmecia (approved):
It's a quite useful feature, LGTM and thanks for the PR again!

PR #609, opened 2026-06-17
fix(docs) Resolve note formatting

czenko:
Closes DOCS-994

Resolves note formatting by following convention and adding an empty line in between.

**Note:** I need help verifying that this change actually worked. Review before approving.


czenko:
Actually, now that I've looking into this more and have more context, I'm going to look into this from the `supabase` side. Since it seems it is correctly formatted in one place and not in another.

PR #612, opened 2026-06-22
fix(mysql_fdw): sanitize MySQL error details

wenerme:
## Summary
- Return sanitized MySQL error details instead of a generic fixed message
- Keep connection/query diagnostics while masking credential-looking values
- Add a focused unit test for the sanitized MySQL error message shape

## Validation
- `git diff --check`
- Conflict marker scan with `rg`
- Not run locally: `cargo`/`rustc` are unavailable in this agent environment; GitHub CI should run the Rust checks/tests


burmecia (approved):
Thanks for the PR, just removed the tests as the `sanitize_error_message` is built-in function and we don't need test it here.

PR #613, opened 2026-06-29
docs: move platform security reference into Contributing docs

spydon:
## What

Moves the root `SECURITY.md` to `docs/contributing/security.md` and adds it to the MkDocs nav under **Contributing → Security Internals**.

References supabase/.github#20

## Why

The root `SECURITY.md` is not a vulnerability-disclosure policy. It's contributor/implementation documentation (WASM checksum verification, credential masking, response size limits) that cites source files like `wasm_fdw.rs` and `utils.rs` and shows how to use internal utilities such as `sanitize_error_message`.

Because GitHub treats any root `SECURITY.md` as the repo's "Security policy", this file was overriding the organization-wide disclosure policy maintained in `supabase/.github`. Moving it:

- keeps the implementation reference on the docs site, next to the other Contributing pages, and
- frees the Security-policy slot so this repo inherits the org-wide default.

The existing user-facing `docs/guides/security.md` (FDW usage, RLS, `security definer`) is unaffected; this is a separate, contributor-focused document.

## Note

Takes effect alongside supabase/.github#20.


burmecia (approved):
thanks @spydon for flagging this!

PR #623, opened 2026-08-09
fix(stripe_fdw): add missing status column to subscriptions table

### LINKED ISSUE #317: Add "status" column in subscriptions stripe to where filter "status" works outside of attrs
I can't search for signatures with status all in the stripe wrapper's foreign table, by default the api only returns signatures with status "active", so we need to have the option to filter by status "all", the code says we can filter by id, costumer and status, but status is inside attrs and doesn't work in reality.

If we do a search on Supabase, it does not return subscriptions with a status other than "active":

select * from stripe.subscriptions where status = 'all';

returns the error:
ERROR: 42703: column "status" does not exist
LINE 1: select * from stripe.subscriptions where status = 'all' limit 100;
^
HINT: Perhaps you meant to reference the column "subscriptions.attrs".

sanskar-soni-9:

Bug fix.

`stripe_fdw` already treats `status` as a pushdown qualifier for `stripe.subscriptions`, and `docs/catalog/stripe.md` recommends filtering subscriptions by `id, customer, price, status` — but the `status` column was never added to the table's column config.
As a result, `where status = 'active'` fails with `ERROR: column "status" does not exist`, even though both the code and the docs advertise it as filterable.

Closes: #317 

## What is the new behavior?

`status text` is added to the `subscriptions` column config and to the example in `docs/catalog/stripe.md`, matching the existing pattern already used for `refunds`.

Also the subscriptions test in `tests.rs` is extended to assert the new column.

## Additional context

`price` has the same missing-column symptom (registered as a pushdown qual at `stripe_fdw.rs:336`, listed in the same docs note) but isn't fixable the same way — `price` lives nested under `items.data[].price` on the Stripe API object, not top-level like `status`, so it needs real flattening logic rather than a column-config addition. Noting it here as a follow-up, not part of this fix.

### DIFF
--- a/docs/catalog/stripe.md
+++ b/docs/catalog/stripe.md
@@ -1010,6 +1010,7 @@ create foreign table stripe.subscriptions (
   currency text,
   current_period_start timestamp,
   current_period_end timestamp,
+  status text,
   attrs jsonb
 )
   server stripe_server
--- a/wrappers/src/fdw/stripe_fdw/README.md
+++ b/wrappers/src/fdw/stripe_fdw/README.md
@@ -10,6 +10,7 @@ This is a foreign data wrapper for [Stripe](https://stripe.com/) developed using
 
 | Version | Date       | Notes                                                |
 | ------- | ---------- | ---------------------------------------------------- |
+| 0.1.13  | 2026-08-10 | Added missing 'status' column to 'subscriptions'     |
 | 0.1.12  | 2025-03-06 | Added import foreign schema support                  |
 | 0.1.11  | 2024-09-20 | Added Meter object                                   |
 | 0.1.10  | 2024-08-26 | Added 'api_key_name' server option                   |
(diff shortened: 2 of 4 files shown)


burmecia (approved):
Thanks for the PR!

sanskar-soni-9:
Hi @burmecia, thanks for merging the PR!

Also as I mentioned in the PR description, `price` has the same problem as the `status` had. But its a special case as `price` is nested inside `items` array.

What do you suggest here?
1. Support `price` as a simple `text` column (populating it with the primary/first item's price) to allow the pushdown filter?
2. Or avoid mapping it to a single column because of multi-item subscriptions, and remove it from the pushdown options?

Let me know your thoughts!

burmecia:
> Hi @burmecia, thanks for merging the PR!
> 
> Also as I mentioned in the PR description, `price` has the same problem as the `status` had. But its a special case as `price` is nested inside `items` array.
> 
> What do you suggest here?
> 
> 1. Support `price` as a simple `text` column (populating it with the primary/first item's price) to allow the pushdown filter?
> 2. Or avoid mapping it to a single column because of multi-item subscriptions, and remove it from the pushdown options?
> 
> Let me know your thoughts!

I think option 2 is better as subscription and price is essentially 1:n relationship, the query for `filtering subscription by price` doesn't seem semantically correct for me.

PR #624, opened 2026-08-11
fix(stripe): remove price pushdown from subscriptions

sanskar-soni-9:

Bug fix,

The `price` parameter is listed as a pushdown filter for `subscriptions`, but is missing from the table columns. Attempting to query or filter by it throws a `column "price" does not exist` error.

## What is the new behavior?

Removed `price` from the `subscriptions` pushdown options and updated the docs.

Since price is nested inside the `items` array and a subscription can have multiple items (a 1:n relationship), we chose to avoid mapping it to a single top-level column and instead remove it from the pushdown options.

## Additional context

This follows the discussion in #623.

### DIFF
--- a/docs/catalog/stripe.md
+++ b/docs/catalog/stripe.md
@@ -1030,7 +1030,6 @@ create foreign table stripe.subscriptions (
 - While any column is allowed in a where clause, it is most efficient to filter by:
       - id
       - customer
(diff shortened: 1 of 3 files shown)


burmecia (approved):
Thanks for the PR!

PR #644, opened 2026-08-31
fix: drop result stream in mysql before pool disconnect to avoid backend hang

Ellba:

Bug fix

Any row-returning query on a MySQL foreign table that the executor stops early (e.g. SELECT * FROM t LIMIT 1) hangs the Postgres backend forever. The backend shows wait_event = NULL and does not respond to pg_cancel_backend / pg_terminate_backend. Because the hung transaction holds a lock on wrappers_fdw_stats, every other FDW query in the database then blocks behind it, until a full restart.

Cause: end_scan() calls pool.disconnect() while self.stream still holds a Conn checked out from that pool. mysql_async::Pool::disconnect() does not resolve until all checked-out connections are dropped, so the backend deadlocks inside rt.block_on(). Fully drained scans (e.g. aggregate pushdown, count(*)) are unaffected, which is why the bug only shows on partial scans.

## What is the new behavior?
end_scan() drops the result stream (self.stream = None) before disconnecting the pool, releasing the checked-out connection so pool.disconnect() resolves immediately. Partial scans now complete and clean up normally.

Added regression test mysql_partial_scan_does_not_hang: LIMIT 1 scans both with and without limit pushdown (the un-pushed path forced via a volatile qual), followed by a full scan on the same foreign table. Against the unfixed code this test hangs indefinitely; with the fix, the full mysql_fdw suite passes 15/15.

Feel free to include screenshots if it includes visual changes.

## Additional context
- Verified with cargo pgrx test --features "mysql_fdw pg17" (PG 17, MySQL 8.0 per .ci/docker-compose-native.yaml).
- Also reproduced standalone against mysql_async 0.36.1, replicating the FDW's structure (current-thread runtime, stream dropped from sync context) to confirm the fix works outside pgrx as well.
- Same bug class as the Iceberg FDW uninterruptible-hang fix

### DIFF
--- a/wrappers/src/fdw/mysql_fdw/mysql_fdw.rs
+++ b/wrappers/src/fdw/mysql_fdw/mysql_fdw.rs
@@ -480,6 +480,11 @@ impl ForeignDataWrapper<MysqlFdwError> for MysqlFdw {
     }
 
     fn end_scan(&mut self) -> MysqlFdwResult<()> {
+        // drop the result stream before disconnecting the pool: a partially
+        // consumed stream still holds a Conn checked out from the pool, and
+        // Pool::disconnect() won't resolve until all such connections are
+        // dropped, so disconnecting first deadlocks the backend
+        self.stream = None;
         self.disconnect_pool()
     }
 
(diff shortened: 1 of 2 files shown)


imor (approved):
Thanks for the PR @Ellba!

PR #645, opened 2026-09-03
fix: double free if a plan is cached and rerun

### LINKED ISSUE #237: Stripe FDW randomly returns no data

imor:
This PR fixes a double free/use after free bug in the core wrappers framework logic. The bug occurred due to an assumption about how Postgres runs the plan and scan phases of a query. Postgres can cache a query once and then execute it repeatedly without ever planning it again. The code in wrappers assumed that Postgres will never cache a plan. This assumption lead to wrappers creating the `FdwState` object at the beginning of the plan phase, and freeing it unconditionally at the end of the first scan phase. If the plan was cached, the new scan phase will then use a freed object and free it again at the end of that phase, and so on.

Since Postgres can cache a plan, it makes a deep copy of the plan via a call to `copyObject` before starting the scan phase. To fix the bug we now ensure that any object created during the plan phase are dropped at the end of that phase and any state we need to carry over from the plan phase to the scan phase we serialize into a `List` via the `fdw_private` member, and reconstitute an `FdwState` struct from this deserialized state. This struct is then dropped at the end of the scan phase. This decoupling of lifetimes of objects in plan vs scan phases is what fixes the bug.

A large portion of the code is this new serialization/deserialization logic, but there were also some other side effects of fixing this bug. Specifically with the above fix we initially started creating two instances of `FdwState` struct (and the Fdw instance) which would break the assumption of FDWs that their instance would only be created ever once, leading to some potential bugs. Less severely, it could also affect performance if and FDW instance performed significant work in their `new()` method, like establishing connections etc. To fix this we also ensure that we now only create an FDW instance during the scan phase. This does mean we had to remove a `&self` argument from certain plan phase functions, but the good news is that this change did not break any of the FDWs.

Also fixes https://github.com/supabase/wrappers/issues/237 because that issue had the same underlying cause.

### DIFF
--- a/supabase-wrappers/src/interface.rs
+++ b/supabase-wrappers/src/interface.rs
@@ -552,12 +552,14 @@ pub struct Qual {
     pub use_or: bool,
     pub param: Option<Param>,
 
-    // Address of the original `pg_sys::Const` (or array `pg_sys::Const`) node this
-    // qual's value was decoded from, if any. Stored as a plain address (not a typed
-    // pointer) so `Qual` stays trivially `Send`-safe; only `scan::get_foreign_plan`
-    // casts it back to a pointer, to embed the original node directly into
-    // `fdw_private` so it survives PostgreSQL's plan-cache `copyObject` correctly.
-    pub(crate) const_node: Option<usize>,
+    // Stores the address of the original const node this qual's value was decoded
+    // from, if any. This is only used during serialization/deserialization to
+    // smuggle the `value` field across the planning and execution phase boundaries
+    // in fdw_private. This ensure the Qual survices the Postgres's plan-cache
+    // copyObject call correctly. It's a usize instead of a *mut pg_sys::Const
+    // to make it Send which is important for certain fdw's like clickhouse which
+    // send it across tokio task boundaries.
+    pub(crate) value_const: Option<usize>,
 }
 
 impl Qual {
--- a/supabase-wrappers/src/interface.rs
+++ b/supabase-wrappers/src/interface.rs
@@ -551,6 +551,13 @@ pub struct Qual {
     pub value: Value,
     pub use_or: bool,
     pub param: Option<Param>,
+
+    // Address of the original `pg_sys::Const` (or array `pg_sys::Const`) node this
+    // qual's value was decoded from, if any. Stored as a plain address (not a typed
+    // pointer) so `Qual` stays trivially `Send`-safe; only `scan::get_foreign_plan`
+    // casts it back to a pointer, to embed the original node directly into
+    // `fdw_private` so it survives PostgreSQL's plan-cache `copyObject` correctly.
+    pub(crate) const_node: Option<usize>,
 }
 
 impl Qual {
@@ -1184,7 +1191,11 @@ pub trait ForeignDataWrapper<E: Into<ErrorReport>> {
         Ok(Vec::new())
     }
 
-    /// Returns a FdwRoutine for the FDW
+    /// The handler function for all foreign data wrappers.
+    ///
+    /// The [`FdwRoutine`] is the same as the `fdw_handler` pseudo-type mentioned in the
+    /// [Postgres documentation](https://www.postgresql.org/docs/current/fdw-functions.html).
+    /// This is the entry point of a foreign table query: the first callback called by Postgres.
     ///
     /// Not to be used directly, use [`wrappers_fdw`](crate::wrappers_fdw) macro instead.
     fn fdw_routine() -> FdwRoutine
--- a/supabase-wrappers/src/interface.rs
+++ b/supabase-wrappers/src/interface.rs
@@ -890,6 +890,12 @@ pub trait ForeignDataWrapper<E: Into<ErrorReport>> {
     /// You can do any initalization in this function, like saving connection
     /// info or API url in an variable, but don't do heavy works like database
     /// connection or API call.
+    ///
+    /// Never called during query planning — [`get_rel_size`](Self::get_rel_size),
+    /// [`supported_aggregates`](Self::supported_aggregates) and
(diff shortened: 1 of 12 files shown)


supabase-wrappers/src/scan.rs
```diff
@@ -175,6 +190,614 @@ unsafe fn drop_fdw_state<E: Into<ErrorReport>, W: ForeignDataWrapper<E>>(
...
+                field,
+                operator,
+                value,
+                use_or,
+                param,
+                value_const: None,
+            })
+        }
+    }
+    unsafe fn read_param(list: &List<*mut c_void>, idx: &mut usize) -> Option<Param> {
```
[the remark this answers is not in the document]
imor: What would cause the list to be trucated/malformed? If something ever corrupts the param, that would the place to fix it correctly not uglying the return value of `read_param` with something like `Option<Option<Param>>`.

wrappers/src/supabase_wrappers_tests.rs
```diff
@@ -250,4 +253,264 @@ mod tests {
...
+            c.update(
+                r#"create foreign table cache_test_table (id bigint) server cache_test_server"#,
+                None,
+                &[],
+            )
+            .unwrap();
+
+            // Use a prepared statement to force plan caching
+            c.update(
+                "prepare cache_test_q as select count(*) from cache_test_table",
```
utkarash2991, line 368: May be we should a test with parameters ex: ```prepare q(bigint) as select * from t where id = $1``` . Execute the same query 6 times so that plan caching kicks in or have the GUC `plan_cache_mode` set to `force_generic_plan` .

supabase-wrappers/src/interface.rs
```diff
@@ -551,6 +551,15 @@ pub struct Qual {
...
     pub param: Option<Param>,
+
+    // Stores the address of the original const node this qual's value was decoded
+    // from, if any. This is only used during serialization/deserialization to
+    // smuggle the `value` field across the planning and execution phase boundaries
+    // in fdw_private. This ensures the Qual survives Postgres' plan-cache
+    // `copyObject` call correctly. It's a usize instead of a *mut pg_sys::Const
+    // to keep `Qual` `Send`, which is important for FDWs like ClickHouse that
+    // move quals across tokio task boundaries.
+    pub(crate) value_const: Option<usize>,
```
utkarash2991, line 562: As we are changing the public facing function signature so I believe we are going to bump the version too once this change is merged right?

wrappers/src/supabase_wrappers_tests.rs
imor: We ensure that cached plan is being used by checking that the value of `PLANNED_CALLS` is 1 and then `NEW_CALLS` is 3, guaranteeing that the query was planned once and run thrice. Did you have concerns with the query plan not being cached?

supabase-wrappers/src/interface.rs
imor: Yes we will bump the version. I usually do that in a separate PR just before releasae.

wrappers/src/supabase_wrappers_tests.rs
utkarash2991: I was just trying to make sure we hit this logic here: https://github.com/postgres/postgres/blob/master/src/backend/utils/cache/plancache.c#L1212 . I dont see any issue but just to be double sure and sake of test completeness.

utkarash2991 (approved):
A nicely done Pr

wrappers/src/supabase_wrappers_tests.rs
imor: That logic is for deciding whether we get a generic or a custom plan, which is independent of whether a plan is cached or not. Since the test uses the `execute` statement, it traces the following code path:

1. Calls `ExecuteQuery` function here: https://github.com/postgres/postgres/blob/c1c5d28f4a2df9d83ca1bcd6bcd6ed8ca065791b/src/backend/commands/prepare.c#L144-L155.
2. Which [calls `GetCachedPlan`](https://github.com/postgres/postgres/blob/c1c5d28f4a2df9d83ca1bcd6bcd6ed8ca065791b/src/backend/commands/prepare.c#L198) defined here: https://github.com/postgres/postgres/blob/c1c5d28f4a2df9d83ca1bcd6bcd6ed8ca065791b/src/backend/utils/cache/plancache.c#L1305-L1308.

This second function returns a `CachedPlan` regardless of whether it's a generic or a custom plan. Anyways we  are testing that we indeed use a cached plan in our test as explained previously.

